PR #5, opened 2020-03-12
Shank 390 api model

jhulbert-va:
https://vasdvp.atlassian.net/browse/SHANK-390

I _believe_ that the Collector-API part of the task is completed by the Facility-API.


facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/error/ApiError.java
```diff
@@ -0,0 +1,27 @@
+package gov.va.api.lighthouse.facilities.api.error;
```
jblefkove-va: I don't think we need two packages... `gov.va.api.lighthouse.facilities.api` seems okay across the board

facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/facilities/FacilitiesSearchResponse.java
```diff
@@ -0,0 +1,47 @@
...
+
+@Value
+@Builder
+@JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY)
+public class FacilitiesSearchResponse {
+  List<Facility> data;
+
+  PageLinks links;
+
+  Metadata meta;
```
jblefkove-va: Since PageLinks and Pagination have been pulled out into their own classes, should Metadata as well?

jhulbert-va: Metadata is different based on the API that you're using. Different requirements.

facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/facilities/Facility.java
```diff
@@ -0,0 +1,352 @@
...
+    String address3;
+
+    String city;
+
+    String state;
+
+    String zip;
+  }
+
+  @Data
```
jblefkove-va: Is this @Data to allow for lazy getters for the unit tests? I'd rather stick to @Value and let the tests be uglier.

jhulbert-va: Yes. Only the classes with lazy getters should have the data annotation. If you find one I labeled incorrectly, let me know.

jblefkove-va: Lazy getters might make things confusing when we're serializing responses and trying to exactly match the current API. I would drop them and do `@Value` everywhere.

jhulbert-va: I added them to match the current API. They prevent the null/empty values from being omitted by Jackson.

jblefkove-va: You should still be able to populate those manually in the unit tests though.

jhulbert-va: :+1:

facilities-api/src/test/java/gov/va/api/lighthouse/facilities/api/facilities/FacilitiesSearchResponseTest.java
```diff
@@ -0,0 +1,190 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.math.BigDecimal;
+import java.time.LocalDate;
+import java.util.ArrayList;
+import java.util.List;
+import lombok.SneakyThrows;
+import org.junit.Test;
+
+public class FacilitiesSearchResponseTest {
```
jblefkove-va: Can we rename the tests to match the sample files they're using? i.e. `SearchByZipGeoJsonTest` in this case?

facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/facilities/GeoFacility.java
```diff
@@ -0,0 +1,91 @@
...
+
+  public enum GeometryType {
+    Point
+  }
+
+  public enum Type {
+    Feature
+  }
+
+  @Data
```
jblefkove-va: 😦

PR #7, opened 2020-03-17
Shank 392 hardcoded collect response

jhulbert-va:
https://vasdvp.atlassian.net/browse/SHANK-392


facilities-collector/pom.xml
```diff
@@ -14,6 +14,10 @@
   <properties>
     <docker.imageName>lighthouse-facilities-collector</docker.imageName>
     <mssql-jdbc.version>7.4.1.jre8</mssql-jdbc.version>
+    <!-- Temporary Decrease. Once /collect is implemented and SneakyThrows isnt required, the
+         contoller should be able to get higher than 75%.
```
jblefkove-va: Why 75%?

facilities-collector/pom.xml
```diff
@@ -29,6 +33,12 @@
       <groupId>org.springframework.boot</groupId>
       <artifactId>spring-boot-starter-data-jpa</artifactId>
     </dependency>
+    <dependency>
+      <groupId>gov.va.api.lighthouse</groupId>
+      <artifactId>facilities-api</artifactId>
+      <version>1.0.6-SNAPSHOT</version>
```
jblefkove-va: Can you do `${project.version}` here instead? And it's not necessary to specify explicit `compile` scope.

facilities-collector/pom.xml
```diff
@@ -29,6 +33,12 @@
       <groupId>org.springframework.boot</groupId>
       <artifactId>spring-boot-starter-data-jpa</artifactId>
     </dependency>
+    <dependency>
+      <groupId>gov.va.api.lighthouse</groupId>
```
jblefkove-va: Keep the `gov.va.*` dependencies together

facilities-collector/spotbugs-excludes.xml
```diff
@@ -1,4 +1,11 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <FindBugsFilter>
     <!-- Exclusions can be defined here and should include a comment on why the finding can be ignored -->
+
+    <!-- Bug is temporary. The hard-coded implementation of the /collect endpoint is only intended to
+         support  the parallel development of the facilities applications other functions. -->
+    <Match>
+        <Bug pattern="OBL_UNSATISFIED_OBLIGATION"/>
```
jblefkove-va: I fear we will forget to remove this later. In the controller, if you open the file resource in a `try-with-resources` block, spotbugs won't complain about this.

facilities-collector/src/main/java/gov/va/api/lighthouse/facilitiescollector/service/controller/collect/CollectController.java
```diff
@@ -0,0 +1,32 @@
+package gov.va.api.lighthouse.facilitiescollector.service.controller.collect;
```
jblefkove-va: I'm inclined to say keep everything in the `gov.va.api.lighthouse.facilitiescollector` package for now

facilities-collector/src/main/java/gov/va/api/lighthouse/facilitiescollector/service/controller/collect/CollectController.java
```diff
@@ -0,0 +1,32 @@
...
+@RestController
+@RequestMapping(
+    value = {"/"},
+    produces = {"application/json"})
+@NoArgsConstructor
+public class CollectController {
+
+  /** Request Mapping for the /collect endpoint. */
+  @SneakyThrows
+  @GetMapping(value = {"/collect"})
```
jblefkove-va: There will eventually be a second endpoint for drive-time-bands, so this feels too generic. How about `/facilities`?

facilities-collector/src/test/java/gov/va/api/lighthouse/facilitiescollector/service/controller/collect/CollectContollerTest.java
```diff
@@ -0,0 +1,13 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import org.junit.Test;
+
+public class CollectContollerTest {
+  @Test
+  public void verifyResponse() {
+    assertThat(new CollectController().collectFacilities())
+        .isExactlyInstanceOf(CollectorFacilitiesResponse.class);
```
jblefkove-va: Seems like this test and the one below should be combined. Invoke `collectFacilities()` on the controller and verify that the result is equal to the sample.

facilities-collector/pom.xml
jhulbert-va: The SneakyThrows annotation isnt being tested.

facilities-collector/src/main/java/gov/va/api/lighthouse/facilitiescollector/service/controller/collect/CollectController.java
jhulbert-va: As in `/collect/facilities` and `/collect/drive-times`? Or were you thinking something different?

facilities-collector/src/test/java/gov/va/api/lighthouse/facilitiescollector/service/controller/collect/CollectContollerTest.java
jhulbert-va: The same used for the hard-coded response is a lot longer than the one I used for the Response test, that's why I did it this way. I also figured that since this was a temporary solution, this test would end up being more comprehensive in its testing of the API.

facilities-collector/src/main/java/gov/va/api/lighthouse/facilitiescollector/service/controller/collect/CollectController.java
jblefkove-va: Just `/facilities` and `/drive-time-bands`

facilities-collector/spotbugs-excludes.xml
jhulbert-va: Still "bugs" out.

PR #10, opened 2020-03-18
SHANK-444 round-trip serialization unit tests

jblefkove-va:
https://vasdvp.atlassian.net/browse/SHANK-444

Configure a Jackson ObjectMapper to match the existing Facilities API serialization behavior for order of fields and emitting nulls. See `JacksonConfigV0` and `SerializeOnlyTest`


facilities-api/pom.xml
```diff
@@ -5,9 +5,20 @@
...
   <version>1.0.9-SNAPSHOT</version>
   <packaging>jar</packaging>
+  <properties>
+    <jackson-datatype-jdk8.version>2.10.3</jackson-datatype-jdk8.version>
+  </properties>
+  <dependencies>
+    <dependency>
+      <groupId>com.fasterxml.jackson.datatype</groupId>
+      <artifactId>jackson-datatype-jdk8</artifactId>
+      <version>${jackson-datatype-jdk8.version}</version>
```
bschofield-va: You should be able to inherit the version through the spring pom's dependencyManagement.
Recommend removing this version node completely.

bschofield-va: i fix!

PR #14, opened 2020-03-19
SHANK-451 Collect state cemeteries

jblefkove-va:
https://vasdvp.atlassian.net/browse/SHANK-451

In addition to the VA cemeteries in ArcGIS, facilities API consumes state cemeteries (not managed by VA) from an XML file maintained by another group. Current facilities API uses a copy of https://www.cem.va.gov/cems/cems.xml in source control; this accesses it directly.


jblefkove-va:
I suggest deferring changes for this PR to https://github.com/department-of-veterans-affairs/lighthouse-facilities/pull/15, which builds on this PR and includes some clean-up for the state cemeteries.

PR #15, opened 2020-03-19
SHANK-399 Collect health facilities

jblefkove-va:
https://vasdvp.atlassian.net/browse/SHANK-399

Includes changes from https://github.com/department-of-veterans-affairs/lighthouse-facilities/pull/14


jblefkove-va:
Good to go

facilities-local-db/src/test/java/gov/va/api/lighthouse/facilitiescollector/Populaterator.java
```diff
@@ -0,0 +1,179 @@
+package gov.va.api.lighthouse.facilitiescollector;
```
bschofield-va: gross... we are going to conflict hard on this file

jblefkove-va: I updated my version with your changes, it should be less painful.

PR #17, opened 2020-03-19
Shank 412 jpa

bschofield-va:
- JPA data model for Facilities
- Liquibase support
- `local-db` now creates a second `facilities` database
- `make-config` now populates facilities properties for a local sql server instance


facilities-collector/src/main/resources/banner.txt
```diff
@@ -1,6 +1,5 @@
 ====
 
     ${Ansi.BLUE}Facilities API ${Ansi.BOLD}${Ansi.YELLOW}CDW ${Ansi.NORMAL}${Ansi.BLUE}${application.version}${Ansi.NORMAL}
-    built in only 4 days
```
jblefkove-va: RIP

bschofield-va:
this is going to need changes to the du

PR #18, opened 2020-03-20
SHANK-398 collect vet centers

jblefkove-va:
https://vasdvp.atlassian.net/browse/SHANK-398


facilities-collector/src/main/java/gov/va/api/lighthouse/facilitiescollector/CdwController.java
```diff
@@ -19,43 +22,77 @@
...
+                        resultSet.getString("NumberOfAppointmentsLinkedToConsult"))
+                    .numberOfLocations(resultSet.getString("NumberOfLocations"))
+                    .avgWaitTimeNew(resultSet.getString("AvgWaitTimeNew"))
+                    .build());
+  }
+
+  @Value
+  @Builder
+  @AllArgsConstructor(access = AccessLevel.PRIVATE)
+  static final class MentalHealthContact {
```
jhulbert-va: Should these be in their own separate classes?

jblefkove-va:
Good to go

facilities-collector/src/test/java/gov/va/api/lighthouse/facilitiescollector/CollectContollerTest.java
```diff
@@ -21,13 +21,23 @@
...
   @SuppressWarnings("unchecked")
   public void verifyResponse() {
     RestTemplate restTemplate = mock(RestTemplate.class);
 
+    ResponseEntity<String> vetCenterResponse = mock(ResponseEntity.class);
+    when(vetCenterResponse.getBody())
+        .thenReturn(
+            JacksonConfig.createMapper().writeValueAsString(ArcGisVetCenters.builder().build()));
+    when(restTemplate.exchange(
+            startsWith("http://arcgis"),
```
jhulbert-va: Anything that uses ArcGIS is going to trample on eachother. Should we be using contains to find the type in the path instead of checking that it begins with the baseurl?

facilities-collector/src/main/java/gov/va/api/lighthouse/facilitiescollector/CdwController.java
jblefkove-va: They can move out if they're needed elsewhere, but I consolidated them since they only exist for these debugging endpoints.

facilities-collector/src/test/java/gov/va/api/lighthouse/facilitiescollector/CollectContollerTest.java
jblefkove-va: Good point

PR #21, opened 2020-03-21
SHANK-458 read with hardcoded response

jblefkove-va:
https://vasdvp.atlassian.net/browse/SHANK-458

Read with a hard-coded response, and error handling to match the existing API. (Examples below. For the last one, the existing API returns a 500 when searching by an unknown facility type, so that is an improvement.)

This PR also includes additional changes (hopefully the last) to the custom Jackson serializers for lazy initialization of certain fields, to make the Facility objects from our collector match the format of the existing API.

```
$ curl -s -H "Accept: application/json+fhir" 'http://localhost:8085/v0/facilities/vha_666'
{"errors":[{"title":"Not acceptable","detail":"The resource could not be returned in the requested format","code":"406","status":"406"}]}

$ curl -s -H "Accept: application/json" 'http://localhost:8085/v0/facilities/vha_666'
{"data":{"id":"vha_666","type":"va_facilities","attributes":{"name":"Sheridan VA Medical Center"...

$ curl -s -H "Accept: application/vnd.geo+json" 'http://localhost:8085/v0/facilities/vha_666'
{"type":"Feature","geometry":{"type":"Point","coordinates":[-106.98488390999995,44.82696953000004]},"properties":{"id":"vha_666",...

$ curl -s 'http://localhost:8085/v0/facilities/vha_666'
{"data":{"id":"vha_666","type":"va_facilities","attributes":{"name":"Sheridan VA Medical Center",...

$ curl -s 'http://localhost:8085/v0/facilities/vha_555'
{"errors":[{"title":"Record not found","detail":"The record identified by vha_555 could not be found","code":"404","status":"404"}]}

$ curl -s 'http://localhost:8085/v0/facilities/xyz_123'
{"errors":[{"title":"Record not found","detail":"The record identified by xyz_123 could not be found","code":"404","status":"404"}]}
```


facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilitiesController.java
```diff
@@ -77,11 +98,36 @@ public void makeSomeGarbage(@PathVariable(name = "sn") String stationNumber) {
...
+  public String readGeoJson(@PathVariable("id") String id) {
+    log.info("Read geo+json facility {}", sanitize(id));
+    GeoFacility geo = GeoFacilityTransformer.builder().facility(readHardcoded(id)).build().toGeoFacility();
+    GeoFacilityReadResponse response =
+        GeoFacilityReadResponse.builder()
+            .type(geo.type())
+            .geometry(geo.geometry())
+            .properties(geo.properties())
+            .build();
+    return JacksonConfigV0.createMapper().writeValueAsString(response);
```
jblefkove-va: @bschofield-va This is suboptimal... do you know a way to configure just this controller class to override the Jackson configuration to use `JacksonConfigV0` instead?

bschofield-va: Not sure how to do it _per controller_ ... but, I think you could do it _per object type_.
1. Created our own implementation of `org.springframework.http.converter.json.MappingJackson2HttpMessageConverter`. Our custom implementation could swap between them. 
2. Create a new `@Configuration` class that supplied our new message converter as a `@Bean`. This the presence of our new mapper _should_ cause the standard one to be disabled.

This would be a gross solution.

Out of curiosity... why can't this V0 mapper be used everywhere? If `JacksonConfigV0` can safely be used for each object, then we can make it `@Configuration` and expose `createMapper` as ` @Bean`. Is there any behavior in this mapper that is contrary to the default one?

bschofield-va: Just briefly looking at it, I _think_ the `JacksonConfigV0` could be simplified to
```
@Configuration
public class JacksonConfigV0 {

  @Bean
  public ObjectMapper createMapper() {
    return JacksonConfig.createMapper().registerModule( customSerializers() );
  }

  private SimpleModel customSerializers() {
    ...
  }
}


```

jblefkove-va: I don't want to use the V0 mapper everywhere because it is full of "ad hoc" rules to mimic the current API. Looking forward to V1 of our API, I want to standardize on the much simpler serialization strategies used by data-query and our other apps.

bschofield-va: Are the rule isolated to types? Or does a single type have different rules in different contexts?

jblefkove-va:
Good to go

PR #28, opened 2020-03-25
SHANK-416 Search by IDs

jblefkove-va:
https://vasdvp.atlassian.net/browse/SHANK-416

This establishes paging behavior. (Paging seems unnecessary for this kind of search, but the existing API offers it.) Pagination is based on how it works in data-query; this expands on the current API's behavior to allow a page size of zero to get an empty response with the total number of results.


facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilitiesController.java
```diff
@@ -27,36 +37,126 @@
...
+    List<FacilityEntity.Pk> pks =
+        Splitter.on(",")
+            .splitToStream(ids)
+            .map(id -> trimToNull(id))
+            .filter(Objects::nonNull)
+            .distinct()
+            .map(
+                id -> {
+                  try {
+                    return FacilityEntity.Pk.fromIdString(id);
```
bschofield-va: Do we want to create `Pk.optionallyFromIdString(id)` that returns `null` or `Optional.empty()` if the ID can't be parsed.

jblefkove-va: 👍

facilities/src/main/java/gov/va/api/lighthouse/facilities/Parameters.java
```diff
@@ -0,0 +1,61 @@
...
+    if (page == null) {
+      return 1;
+    }
+    return Integer.parseInt(page);
+  }
+
+  static int perPageOf(MultiValueMap<String, String> parameters) {
+    String perPage = parameters.getFirst("per_page");
+    if (perPage == null) {
+      return 10;
```
jhulbert-va: Should this be moved to `application.properties` so that the default can be changed easily?

jhulbert-va:
I know they end up being tested in the `FacilitiesControllerTest`, but should there be separate tests for `PageLinks` and `Parameters`?

facilities/src/main/java/gov/va/api/lighthouse/facilities/Parameters.java
jblefkove-va: We can turn this into an exception instead since paging parameters should always be provided.

jblefkove-va:
Feedback addressed

PR #34, opened 2020-03-27
SHANK-421 search by bounding box

jblefkove-va:
https://vasdvp.atlassian.net/browse/SHANK-421

Filtering behavior is intentionally less restrictive than the Rails API. (You can filter by `type` and `services` independently. In the previous API, services filtering can only be done in addition to type filtering.)


facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/v0/Facility.java
```diff
@@ -75,7 +75,6 @@
   }
 
   public enum OtherService implements ServiceType {
-    @JsonProperty("Online Scheduling")
```
jhulbert-va: Why is that being removed? Isn't there supposed to be a space in the output from the api?

facilities/pom.xml
```diff
@@ -14,7 +14,7 @@
   <properties>
     <docker.imageName>lighthouse-facilities</docker.imageName>
     <guava.version>28.2-jre</guava.version>
-    <jacoco.coverage>0.91</jacoco.coverage>
+    <jacoco.coverage>0.80</jacoco.coverage>
```
jhulbert-va: ?

facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilitiesController.java
```diff
@@ -127,6 +255,39 @@ public GeoFacilitiesResponse geoFacilitiesByIds(
...
+                    .build())
+            .totalEntries(entities.size())
+            .build();
+    return FacilitiesResponse.builder()
+        .data(page(entities, page, perPage).stream().map(e -> facility(e)).collect(toList()))
+        .links(linker.links())
+        .meta(
+            FacilitiesResponse.FacilitiesMetadata.builder()
+                .pagination(linker.pagination())
+                .distances(emptyList())
```
jhulbert-va: Why not use `@Builder.Default` to set this in the model?

facilities/pom.xml
jhulbert-va: Nvm I see the note above about unit tests.

facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilitiesController.java
jblefkove-va: I don't want to default it the API class itself, as that is not consistent with how the others work. But you're correct that it can be done in one place... this kludgey "match the existing api" behavior can be handled in the serializers.

facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/v0/Facility.java
jblefkove-va: I suspect it's supposed to be `OnlineScheduling` and the Swagger typo wasn't caught because `other` services have never been populated. (There is nothing in the existing API for it.) The existing API also doesn't let you filter by this (it restricts to health or benefits services explicitly), but if for some reason someone does filter on it in our API, it should be `OnlineScheduling`.

jblefkove-va: I reached out for clarification.

jblefkove-va:
Good to go

jblefkove-va:
Think I found a bug, don't merge

jblefkove-va:
All good

bschofield-va:
Now that integration tests work, are there ITs we need to turn on here?

bschofield-va:
```
[ERROR]   FacilitiesIT.searchByBoundingBoxWithServicesMissingType:90 1 expectation failed.
Expected status code <400> but was <200>.
```

bschofield-va:
hah!  I just failed to push because I was making the exact change.

jblefkove-va:
Great minds think alike

PR #36, opened 2020-03-30
SHANK-427 All as CSV

jblefkove-va:
https://vasdvp.atlassian.net/browse/SHANK-427


facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilitiesController.java
```diff
@@ -154,17 +154,36 @@ private static GeoFacility geoFacility(Facility facility) {
...
         .build();
   }
 
+  /** Get all facilities as CSV. */
+  @SneakyThrows
+  @GetMapping(value = "/facilities/all", produces = "text/csv")
+  public String allCsv() {
+    List<List<String>> rows =
+        Streams.stream(facilityRepository.findAll())
+            .parallel()
```
jblefkove-va: Converting each facility to `List<String>` for the CSV is surprisingly slow. In serial the whole call takes about 16 seconds on my laptop. Adding parallel reduces that to a few seconds. (This is probably okay as I suspect this is a rarely-used endpoint. But if that turns out not to be the case, we should look into caching this.)

bschofield-va: Would be curious to know how much faster it  might be if you used DTO projections instead of the JPA entities themselves. It's not urgent, but we may consider going back after functionality is complete to improve performance.

bschofield-va: Current API does it 2 seconds, fyi

PR #38, opened 2020-03-31
SHANK-464 Use stop codes for podiatry and nutrition

jblefkove-va:
https://vasdvp.atlassian.net/browse/SHANK-464

(Also includes changes from https://github.com/department-of-veterans-affairs/lighthouse-facilities/pull/37)


jblefkove-va:
There is some ambiguity on how a service's overall wait time should be represented when there are multiple stop codes for it. I've added a follow-up story, https://vasdvp.atlassian.net/browse/SHANK-503, and will remove that functionality from this PR.

facilities-collector/spotbugs-excludes.xml
```diff
@@ -3,10 +3,11 @@
...
   <!-- Allow null Boolean in facility model. -->
   <Match>
-    <Or>
-      <Class name="HealthTransformer"/>
-      <Method name="mobile"/>
-    </Or>
+    <Method name="mobile"/>
     <Bug pattern="NP_BOOLEAN_RETURN_NULL"/>
   </Match>
+  <!-- Debugging class -->
```
bschofield-va, line 9: What issue was being flagged here? I suspect that Fortify will also complain and we won't be able to explain it away so easily.

facilities-collector/src/main/java/gov/va/api/lighthouse/facilitiescollector/CdwController.java
```diff
@@ -22,74 +26,40 @@
 public class CdwController {
   private final JdbcTemplate jdbc;
 
+  @SneakyThrows
+  private List<Map<String, String>> allResults(String sql) {
+    List<Map<String, String>> results = new ArrayList<>();
+    try (Connection connection = jdbc.getDataSource().getConnection()) {
+      try (PreparedStatement statement = connection.prepareStatement(sql)) {
+        try (ResultSet resultSet = statement.executeQuery()) {
+          ResultSetMetaData rsmd = resultSet.getMetaData();
```
bschofield-va, line 35: I don't know if this helps readability, but you can group try-with-resources.
```
try(
   var connection = jdbc.getDataSource().getConnection();
   var statement = connection.prepareStatement(sql);
   var resultSet = statement.executeQuery()
) {
 // ... stuff ...
}
```

facilities-collector/src/main/java/gov/va/api/lighthouse/facilitiescollector/HealthsCollector.java
```diff
@@ -86,35 +85,66 @@ static void putMentalHealthContact(ResultSet resultSet, Map<String, String> map)
...
     String id = "vha_" + stationNumber;
     if (map.containsKey(id)) {
       log.warn("Duplicate mental health contact for station number {}", sanitize(stationNumber));
     }
-
     map.put(id, phone);
   }
 
+  @SneakyThrows
+  static void putStopCode(ResultSet resultSet, ListMultimap<String, StopCode> map) {
```
bschofield-va, line 109: Let's add a little logging in here so we know how long this is taking

facilities-collector/src/main/java/gov/va/api/lighthouse/facilitiescollector/HealthTransformer.java
```diff
@@ -376,10 +375,20 @@ private BigDecimal satisfactionScore(String type) {
     if (dentalServiceFacilityIds.contains(id())) {
       services.add(Facility.HealthService.DentalServices);
     }
+    if (stopCodes().stream().anyMatch(sc -> StopCode.NUTRITION.contains(trimToEmpty(sc.code())))) {
```
bschofield-va, line 378: Nothing wrong with this, but is this more readable? But are we better making a generic mechanism for custom stop-code-to-service-type mapping?
```
services.addAll(
   stopCodes().stream().map(StopCode::additionalServiceType).filter(Object::nonNull).collect(toList()))
```

anyway, I'm good to proceed with what's here, but if we have to add another special case, I want to replace the if/thens with a generic mechanism.

facilities-collector/spotbugs-excludes.xml
jblefkove-va: Complaining simultaneously about potential NPEs and unnecessary null-checks. I'll take another look.

facilities-collector/src/main/java/gov/va/api/lighthouse/facilitiescollector/HealthsCollector.java
jblefkove-va: @jhulbert-va pointed out last sprint that we don't get the benefits of the automatic logging on the individual collector classes, because they are not Spring components. Is that something we should look into refactoring?

facilities-collector/src/main/java/gov/va/api/lighthouse/facilitiescollector/HealthTransformer.java
jblefkove-va: Agreed, piecemeal approach is okay for now, we're only using three services types. (Podiatry and nutrition, and dental services soon.) If we do get around to replacing access-to-care with the stop code data wholesale, we will need to reevaluate.

facilities-collector/src/main/java/gov/va/api/lighthouse/facilitiescollector/HealthsCollector.java
bschofield-va: No, I don't think so. The `@Loggable` annotation when applied explicitly (or implicitly for @RestController) only logs "level 0" by default anyway. So if this was componentized, you'd only see those statements on errors. I think just adding old fashion logging is enough.

jblefkove-va:
Feedback addressed

PR #39, opened 2020-03-31
Shank 418 search by zip

eclendenning-va:


eclendenning-va:
I wanted to point out the fact i am using `criteriaBuilder.like` instead of `criteriaBuilder.equal` to deal with the 5 character zips as well as the 10 character zips. I am unsure if this is the correct way to do this so I wanted some extra scrutiny on this part.

jblefkove-va:
> I wanted to point out the fact i am using `criteriaBuilder.like` instead of `criteriaBuilder.equal` to deal with the 5 character zips as well as the 10 character zips. I am unsure if this is the correct way to do this so I wanted some extra scrutiny on this part.

I think regular equality is sufficient. The zip search parameter gets truncated to five characters, and all the facility entities are populated with five-character zips: https://github.com/department-of-veterans-affairs/lighthouse-facilities/blob/master/facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalManagementController.java#L92

PR #45, opened 2020-04-01
SHANK-420 search by lat long

jblefkove-va:
https://vasdvp.atlassian.net/browse/SHANK-420


jblefkove-va:
Note on distances: The Rails API uses PostGIS to compute the distance, while the haversine formula is used here. So the results don't exactly match. But, if you Google "distance between GPS coordinates", one of the top results is [this page](http://boulter.com/gps/distance/); I've compared a few of our calculations there, and they exactly match. (At least, up to two decimal places.) So I'm confident that the algorithm was implemented correctly.

Furthermore, I suspect that this distances feature is a legacy holdover that our consumers aren't going to use anyway. (This is "crow flies" distance. If a consumer has business rules that depend on the distance between a facility and some other point, they should be using the `/nearby` endpoint instead.)

facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilitiesController.java
```diff
@@ -408,6 +469,54 @@ public FacilitiesResponse jsonFacilitiesByIds(
...
+      @RequestParam(value = "lat") BigDecimal latitude,
+      @RequestParam(value = "long") BigDecimal longitude,
+      @RequestParam(value = "type", required = false) String type,
+      @RequestParam(value = "services[]", required = false) List<String> services,
+      @RequestParam(value = "page", defaultValue = "1") @Min(1) int page,
+      @RequestParam(value = "per_page", defaultValue = "10") @Min(0) int perPage) {
+    List<DistanceEntity> entities = entitiesByLatLong(longitude, latitude, type, services);
+    PageLinker linker =
+        PageLinker.builder()
+            .url(baseUrl + "v0/facilities")
```
jhulbert-va, line 487: Dont forget the base path here.

jblefkove-va: Whoops

jblefkove-va:
Good to go

PR #46, opened 2020-04-01
SHANK-510 Standardize RestTemplate for *.va.gov

jblefkove-va:
https://vasdvp.atlassian.net/browse/SHANK-510


jhulbert-va:
:shipit:

PR #48, opened 2020-04-01
add basePath

eclendenning-va:
https://github.com/department-of-veterans-affairs/lighthouse-facilities-deployment/pull/14


make-configs.sh
```diff
@@ -47,6 +47,7 @@ configValue facilities-collector $PROFILE va-arc-gis.url 'http://localhost:8666'
 makeConfig facilities $PROFILE
 configValue facilities $PROFILE facilities-collector.url 'http://localhost:8080'
 configValue facilities $PROFILE facilities.url 'http://localhost:8085'
+configValue facilities $PROFILE facilities.base-path 'va_facilities'
```
jhulbert-va, line 50: Locally, this would just be `/` the `/va_facilities` is stripped by the ingress.

facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilitiesController.java
```diff
@@ -358,7 +362,7 @@ public FacilitiesResponse jsonFacilitiesByBoundingBox(
     List<FacilityEntity> entities = entitiesByBoundingBox(bbox, type, services);
     PageLinker linker =
         PageLinker.builder()
-            .url(baseUrl + "v0/facilities")
+            .url(baseUrl + basePath + "v0/facilities")
```
jhulbert-va, line 365: @jblefkove-va do you think this should take place in `PageLinker` or do you think its fine the way it is?
By that I mean give the linker `v0/facilities` and have baseUrl and basePath loaded and aggregated in the linker.

bschofield-va: Either way is fine with me, but since PageLinker isn't a spring component, I'd just assume _this_ class make the decision to combine `baseUrl` + `basePath` since it is the consumer of the configuration properties.

jhulbert-va: Sounds good to me. Approved.

facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilitiesController.java
```diff
@@ -86,13 +86,17 @@
...
   FacilitiesController(
       @Autowired FacilityRepository facilityRepository,
       @Autowired DriveTimeBandRepository driveTimeBandRepository,
-      @Value("${facilities.url}") String baseUrl) {
+      @Value("${facilities.url}") String baseUrl,
+      @Value("${facilities.base-path}") String basePath) {
     this.facilityRepository = facilityRepository;
     this.driveTimeBandRepository = driveTimeBandRepository;
     this.baseUrl = baseUrl.endsWith("/") ? baseUrl : baseUrl + "/";
+    this.basePath = basePath.endsWith("/") ? basePath : basePath + "/";
```
jblefkove-va, line 99: Sorry, nitpick... if `basePath` is empty with a trailing slash, the links have a double slash: `http://localhost:8085//v0/facilities?state=FL`

While this is still a valid link, it looks silly. For basePath we should make it empty if it's equal to `/`.

PR #52, opened 2020-04-02
SHANK-469 delete stale facilities on reload

jblefkove-va:
https://vasdvp.atlassian.net/browse/SHANK-469


facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilityManagementController.java
```diff
@@ -97,16 +104,45 @@ private void createNewEntity(ReloadResponse response, Pk pk, Facility facility)
...
 
+  private void deleteStaleFacilities(
+      ReloadResponse response, CollectorFacilitiesResponse collectedFacilities) {
+    Set<Pk> newIds =
+        collectedFacilities.facilities().stream()
+            .map(f -> FacilityEntity.Pk.optionalFromIdString(f.id()).orElse(null))
+            .filter(Objects::nonNull)
+            .collect(Collectors.toCollection(LinkedHashSet::new));
+    Set<FacilityEntity.Pk> oldIds =
+        Streams.stream(facilityIdRepository.findAll())
```
jblefkove-va, line 115: Using an entity here that represents only the facility ID produces some modest speedup on my laptop. (About 500 ms vs 100 ms for all of `deleteStaleFacilities()`)

PR #57, opened 2020-04-04
SHANK-409 nearby lat long

jblefkove-va:
https://vasdvp.atlassian.net/browse/SHANK-409


facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilitiesController.java
```diff
@@ -152,6 +156,31 @@ private static double haversine(@NonNull FacilityEntity entity, double lng, doub
...
+    }
+    shape.closePath();
+    return shape;
+  }
+
+  private static Integer validateDriveTime(Integer val) {
+    if (val == null) {
+      return null;
+    }
+    if (val <= 9 || val >= 91 || val % 10 != 0) {
```
jhulbert-va, line 178: Would it be better here to just have a list of accepted values (10, 20, 30, etc.) and checking that the value is in that list?

facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilitiesController.java
```diff
@@ -587,13 +660,113 @@ public FacilitiesResponse jsonFacilitiesByZip(
...
+                .vaFacility(
+                    NearbyResponse.VaFacility.builder()
+                        .links(
+                            NearbyResponse.Links.builder()
+                                .related(linkerUrl + "facilities/" + entity.facilityId())
+                                .build())
+                        .build())
+                .build())
+        .build();
+  }
```
jhulbert-va, line 721: Should this be in its own transformer?

facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilitiesController.java
jblefkove-va: Updated

facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilitiesController.java
jblefkove-va: If it were more complicated with lots of null checks, etc, I would extract it. But since it's just this single method, I'm inclined to leave it.

facilities/src/main/java/gov/va/api/lighthouse/facilities/DriveTimeBandRepository.java
```diff
@@ -10,8 +21,38 @@
...
+    public Predicate toPredicate(
+        Root<DriveTimeBandEntity> root,
+        CriteriaQuery<?> criteriaQuery,
+        CriteriaBuilder criteriaBuilder) {
+      List<Predicate> predicates = new ArrayList<>(5);
+
+      predicates.add(criteriaBuilder.lessThanOrEqualTo(root.get("minLongitude"), longitude));
+      predicates.add(criteriaBuilder.greaterThanOrEqualTo(root.get("maxLongitude"), longitude));
+      predicates.add(criteriaBuilder.lessThanOrEqualTo(root.get("minLatitude"), latitude));
+      predicates.add(criteriaBuilder.greaterThanOrEqualTo(root.get("maxLatitude"), latitude));
```
jhulbert-va, line 48: Reading this feels a little weird. Shouldn't the values be less than the max and greater than the min?

eclendenning-va: Its checking if the min is less than or equal to the value or if the max is greater than or equal to the the value. it looks weird at first but the criteria is correct,

jblefkove-va: No, it's like "first argument, method name, second argument". So `criteriaBuilder.lessThanOrEqualTo(root.get("minLongitude"), longitude)` is `entity.minLongitude <= longitude`. (We want to verify that the lat/long search parameters are within the bounding box around the band.)

jhulbert-va: Gotcha. I was just reading it weird then.

PR #58, opened 2020-04-06
Shank 512 cms part 1

bschofield-va:
This adds model changes and introduces the concept of FacilityOverlay. The Overlay will eventually smerge CMS data into the final Facility, but for now just contains logic to ensure the operation status is populated.


facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/v0/GeoFacility.java
```diff
@@ -91,6 +91,11 @@
     @JsonProperty("active_status")
     Facility.ActiveStatus activeStatus;
 
+    @Valid
+    @NotNull
+    @JsonProperty("operating_status")
```
jhulbert-va, line 96: Should this also be `required = true`?

PR #70, opened 2020-04-11
SHANK-531 /all additional performance

jblefkove-va:
https://vasdvp.atlassian.net/browse/SHANK-531


bschofield-va:
Before we merge this, I'd like to try Jackson Afterburner to measure the effect it has.

jblefkove-va:
Branch for afterburner: https://github.com/department-of-veterans-affairs/lighthouse-facilities/compare/SHANK-000-afterburner

There is no additional configuration beyond adding the module? I tried it on master and on this branch, did not observe a significant performance difference in either case.

jblefkove-va:
I also tried an experiment using `writeValueAsBytes` instead of `writeValueAsString`. That is, converting the facilities to bytes in parallel, then stitching them together in one big byte array. Surprisingly this was slower than the `StringBuilder` approach. We still have the option of caching the whole string response, but I think this is good-to-go.

PR #74, opened 2020-04-16
SHANK-537 investigate valid parameter combinations

jblefkove-va:
https://vasdvp.atlassian.net/browse/SHANK-537

Align parameter combinations with the behavior of the previous API:
- `bbox[]`, `lat/long`, `state`, `zip` are mutually exclusive
- `ids` can be specified by itself, or as a filter parameter `bbox[]`. It will be accepted (but ignored) for the others.
- `type` and `services[]` can be specified as filter parameters for any of the four location-based paramters. They will be accepted (but ignored) for `ids`.


bschofield-va:
I wish there was clear list of valid parameter combinations. The description is better, but I think we could provide a concise list. This is my interpretation of valid combinations
`*` specified
`?` optionally specified
`-` not specified

```
ids      * ? ? ? ?
zip      - * - - -
state    - - * - -
lat      - - - * -
long     - - - * -
bbo-     - - - - *
type     ? ? ? ? ?
services ? ? ? ? ?
page     ? ? ? ? ?
per_page ? ? ? ? ?
```

Perhaps something like this?

> **Valid parameter combinations**
> You may optionally specify `type`, `services[]`, `page` and `page_size` with all queries.
> You must specify one of the following parameter combinations
> - `ids`
> - `zip`, optionally with `ids`
> - `state`, optionally with `ids`
> - `lat`, `long`, optionally with `ids`
> - `bbox[]`, optionally with `ids`

jblefkove-va:
How about:

> **Valid parameter combinations**
> You may optionally specify `page` and `page_size` with all queries.
> You must specify one of the following parameter combinations:
> 
> * `ids`
> * `zip`, optionally with `type` and `services[]`
> * `state`, optionally with `type` and `services[]`
> * `lat` and `long`, optionally with `ids`, `type`, and `services[]`
> * `bbox[]`, optionally with `type` and `services[]`

facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/FacilitiesAllApi.java
```diff
@@ -18,11 +18,11 @@
       tags = {"facilities"},
       summary = "Bulk download of all available facilities",
       description =
-          "Retrieve all available facilities in a single operation, formatted as either a GeoJSON\n"
+          "Retrieve all available facilities in a single operation, formatted as either a GeoJSON "
```
jhulbert-va, line 21: The new lines need to stay. It formats it on the dev-portal. Without them, things end up looking weird.

jblefkove-va: My understanding is that as long as it looks okay on editor.swagger.io, it will be fine on the developer portal as well. I will look into running developer portal locally to verify.

jblefkove-va:
Verified Swagger changes on the developer portal locally:
image

PR #78, opened 2020-04-21
Reduce logging spam on test failure

bschofield-va:


facilities-tests/pom.xml
```diff
@@ -4,7 +4,7 @@
   <parent>
     <groupId>gov.va.api.health</groupId>
     <artifactId>test-starter</artifactId>
-    <version>6.0.4</version>
+    <version>6.0.5-SNAPSHOT</version>
```
jhulbert-va, line 7: Remove SNAPSHOT before merge.

PR #79, opened 2020-04-21
Better logging while collecting

bschofield-va:
Reload response now includes a total number of facilities collected
Better logging of times and data while collecting


facilities/src/main/java/gov/va/api/lighthouse/facilities/ReloadResponse.java
```diff
@@ -13,6 +13,7 @@
 @Builder
 @Data
 public class ReloadResponse {
+  private int totalFacilities;
```
jblefkove-va, line 16: Does `totalFacilities` always equal `facilitiesCreated.size() + facilitiesUpdated.size()` ?

bschofield-va: In theory, but not guaranteed. If we encounter a facility with an ID we cannot parse, then it would be off.

PR #84, opened 2020-04-27
Shank 545 covid 19 info

jhulbert-va:
# BLOCKED
### https://lighthouseva.slack.com/archives/CJ162GDDJ/p1588013276239500

https://vasdvp.atlassian.net/browse/SHANK-545


facilities-collector/src/main/java/gov/va/api/lighthouse/facilitiescollector/AccessToCareCovid19Entry.java
```diff
@@ -0,0 +1,59 @@
...
+  @JsonIgnore
+  @JsonUnwrapped
+  @Getter(onMethod_ = @JsonAnyGetter)
+  Map<String, Object> summary;
+
+  /**
+   * Lombok doesnt play well with the JsonAnySetter annotation, so a custom builder needs to exist
+   * that deals with this.
+   */
+  @JsonPOJOBuilder(withPrefix = "")
```
ilaflamme-va, line 42: This seems odd to me. have we not handled this before?

jhulbert-va: I dont think we've ever used the `JsonAnySetter` annotation before, no.

facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/v0/Facility.java
```diff
@@ -427,4 +429,13 @@ public HoursBuilder wed(String val) {
     @JsonProperty("effective_date")
     LocalDate effectiveDate;
   }
+
+  @Data
+  @Builder
+  @JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY)
+  public static final class Covid19 {
+    Integer confirmedCases;
```
jblefkove-va, line 437: `@Min(0)` ?

facilities-collector/src/main/java/gov/va/api/lighthouse/facilitiescollector/HealthsCollector.java
```diff
@@ -228,6 +230,40 @@ private ArcGisHealths loadArcGis() {
...
+    String response =
+        insecureRestTemplate
+            .exchange(url, HttpMethod.GET, new HttpEntity<>(new HttpHeaders()), String.class)
+            .getBody();
+    List<AccessToCareCovid19Entry> entries =
+        JacksonConfig.createMapper()
+            .readValue(response, new TypeReference<List<AccessToCareCovid19Entry>>() {});
+    ListMultimap<String, AccessToCareCovid19Entry> map = ArrayListMultimap.create();
+    for (AccessToCareCovid19Entry entry : entries) {
+      String facilityId = entry.facility();
```
jblefkove-va, line 249: This is really `stationId`. The `facilityId` is `vha_stationId`.

facilities-collector/src/main/java/gov/va/api/lighthouse/facilitiescollector/HealthsCollector.java
```diff
@@ -228,6 +230,40 @@ private ArcGisHealths loadArcGis() {
...
+            .build()
+            .toUriString();
+    String response =
+        insecureRestTemplate
+            .exchange(url, HttpMethod.GET, new HttpEntity<>(new HttpHeaders()), String.class)
+            .getBody();
+    List<AccessToCareCovid19Entry> entries =
+        JacksonConfig.createMapper()
+            .readValue(response, new TypeReference<List<AccessToCareCovid19Entry>>() {});
+    ListMultimap<String, AccessToCareCovid19Entry> map = ArrayListMultimap.create();
```
jblefkove-va, line 247: A regular map should be sufficient; we only expect one entry per facility. (For the AccessToCare wait times and satisfaction scores, a multimap is necessary since those are one-entry-per-service; we expect multiple for each facility.)

jblefkove-va:
Does this include Swagger updates?

facilities-collector/src/main/java/gov/va/api/lighthouse/facilitiescollector/AccessToCareCovid19Entry.java
```diff
@@ -0,0 +1,59 @@
...
+  @JsonProperty("FacilityLongitude")
+  BigDecimal longitude;
+
+  @JsonProperty("Facility")
+  String facility;
+
+  @JsonIgnore
+  @JsonUnwrapped
+  @Getter(onMethod_ = @JsonAnyGetter)
+  Map<String, Object> summary;
```
jblefkove-va, line 36: This inner map is going to cause a lot of headaches. Replace this with explicit `confirmedCases` and `deaths` `Integer` fields. Add logic to the `@JsonAnySetter` to populate `confirmedCases` if the key contains `confirmed`, etc.

Similarly, you could rename the `facility` field to `stationId` and move the logic for extracting the station ID from the `Facility` key into the `@JsonAnySetter` method as well.

jhulbert-va:
> Does this include Swagger updates?

It should. When covid19 was added to `Facility.java` in the `facilities-api` module, swagger should have pulled that in. What other updates were you thinking?

jblefkove-va:
> > Does this include Swagger updates?
> 
> It should. When covid19 was added to `Facility.java` in the `facilities-api` module, swagger should have pulled that in. What other updates were you thinking?

`@Schema` descriptions and examples

facilities-collector/src/main/java/gov/va/api/lighthouse/facilitiescollector/AccessToCareCovid19Entry.java
```diff
@@ -0,0 +1,79 @@
...
+        log.info("Collect access-to-care-covid19: Invalid station id ({}).", stnId);
+      }
+      return stnId;
+    }
+
+    /** Uses Jacksons JsonAnySetter annotation to map any unknown values to a map. */
+    @JsonAnySetter
+    public AccessToCareCovid19EntryBuilder saveUnknown(String key, Object value) {
+      if (containsIgnoreCase(key, "confirmed")) {
+        confirmedCases = isNumeric(value.toString()) ? (Integer) value : null;
```
bschofield-va, line 72: Casting without verifying type is unsafe.
We cannot trust these values.
- It could be `null` --> NPE
- It could be String --> ClassCastException
- It could be Long --> ClassCastException
- It could be Double --> ClassCastException

We should handle these cases
- `null` --> `null`
- `String` --> Assume integer, `Integer.parseInt( value )` + catch/handle exception
- `java.lang.Number` --> `value.intValue()`
- _mystery value_ --> handle exception

Exceptions should be handled per: https://lighthouseva.slack.com/archives/CJ162GDDJ/p1588013063237800

facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/v0/Facility.java
```diff
@@ -435,9 +435,11 @@ public HoursBuilder wed(String val) {
   @Builder
   @JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY)
   public static final class Covid19 {
+    @Schema(description = "Number of confirmed Covid-19 cases at a facility.", example = "46")
     @Min(0)
     Integer confirmedCases;
 
+    @Schema(description = "Number of confirmed deaths by Covid-19 at a facility.", example = "4")
```
eclendenning-va, line 442: Nit : "Number of confirmed deaths by Covid-19 at a facility." Seems weird to me,  how about  "Number of Covid-19 related deaths at a facility."  ?

facilities-collector/src/main/java/gov/va/api/lighthouse/facilitiescollector/AccessToCareCovid19Entry.java
jhulbert-va: `isNumeric` will test for all those cases listed. If we want different behavior for each, I can update.

facilities-collector/src/main/java/gov/va/api/lighthouse/facilitiescollector/HealthsCollector.java
```diff
@@ -228,6 +231,31 @@ private ArcGisHealths loadArcGis() {
...
+  @SneakyThrows
+  private Map<String, AccessToCareCovid19Entry> loadCovid19AccessToCare() {
+    final Stopwatch watch = Stopwatch.createStarted();
+    String url =
+        UriComponentsBuilder.fromHttpUrl(atcBaseUrl + "covid19/vacovid19summary.json")
+            .build()
+            .toUriString();
+    String response =
+        insecureRestTemplate
+            .exchange(url, HttpMethod.GET, new HttpEntity<>(new HttpHeaders()), String.class)
```
bschofield-va, line 243: If this fails to download, it will kill the whole collection. We need to handle this as a recoverable error.

facilities-collector/src/main/java/gov/va/api/lighthouse/facilitiescollector/HealthsCollector.java
```diff
@@ -228,6 +231,31 @@ private ArcGisHealths loadArcGis() {
...
+            .getBody();
+    List<AccessToCareCovid19Entry> entries =
+        JacksonConfig.createMapper()
+            .readValue(response, new TypeReference<List<AccessToCareCovid19Entry>>() {});
+    Map<String, AccessToCareCovid19Entry> map = new HashMap<>();
```
bschofield-va, line 248: You don't have to use this, but if you're not familiar with collecting to maps, check out:
```
entries.stream().collect(Collectors.toMap( keyFunction, valueFunction )
```

facilities-collector/src/main/java/gov/va/api/lighthouse/facilitiescollector/HealthsCollector.java
```diff
@@ -228,6 +231,31 @@ private ArcGisHealths loadArcGis() {
...
+    final Stopwatch watch = Stopwatch.createStarted();
+    String url =
+        UriComponentsBuilder.fromHttpUrl(atcBaseUrl + "covid19/vacovid19summary.json")
+            .build()
+            .toUriString();
+    String response =
+        insecureRestTemplate
+            .exchange(url, HttpMethod.GET, new HttpEntity<>(new HttpHeaders()), String.class)
+            .getBody();
+    List<AccessToCareCovid19Entry> entries =
```
bschofield-va, line 245: Failure to parse will cause an exception and abort collection. We need to handle this as a recoverable error.

facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/v0/Facility.java
```diff
@@ -190,6 +204,8 @@
     @NotNull
     @JsonProperty(value = "operating_status", required = true)
     OperatingStatus operatingStatus;
+
+    @Valid Covid19 covid19;
```
bschofield-va, line 208: This should have some `@Schema( description )`

facilities-collector/src/main/java/gov/va/api/lighthouse/facilitiescollector/AccessToCareCovid19Entry.java
```diff
@@ -0,0 +1,83 @@
...
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Value;
+import lombok.extern.slf4j.Slf4j;
+
+@Value
+@Builder
+@AllArgsConstructor(access = AccessLevel.PRIVATE)
+@JsonDeserialize(builder = AccessToCareCovid19Entry.AccessToCareCovid19EntryBuilder.class)
+public class AccessToCareCovid19Entry {
```
bschofield-va, line 22: I'd like to see some unit tests verifying this relatively complicated deserialization magic to verify parsing JSON we understand and do not understand.

facilities-collector/src/main/java/gov/va/api/lighthouse/facilitiescollector/CollectorHealthController.java
```diff
@@ -118,7 +130,7 @@ public void clearCacheScheduler() {
             .parallelStream()
             .map(p -> testDownstreamHealth(p.getFirst(), p.getSecond()))
             .collect(Collectors.toList());
-
+    downstreamServices.add(testAtcCovid19Health());
```
bschofield-va, line 133: Probably worth a comment as to why this is outside of the parallelization and special
... also .. let's make it parallel. You'll need to change what Pair is 
Instead of parallelize a list of suppliers that return a Health.
5 will look the same, 1 will be special.

```
Supplier<Health> basicHealthCheck( String name, String url ) {
   return ()-> testDownstreamHealth(name,url);
}
downstreamServices.add( basicHealthCheck( "Access to Care", ... url ...) );
downstreamServices.add( fancyBoiHealthCheck( this::testAtcCovid19Health() );
downstreamServices.parallelStream().map( f -> f.get() ).collect( toList() )
```

facilities-collector/src/main/java/gov/va/api/lighthouse/facilitiescollector/AccessToCareCovid19Entry.java
```diff
@@ -0,0 +1,83 @@
...
+import lombok.AccessLevel;
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Value;
+import lombok.extern.slf4j.Slf4j;
+
+@Value
+@Builder
+@AllArgsConstructor(access = AccessLevel.PRIVATE)
+@JsonDeserialize(builder = AccessToCareCovid19Entry.AccessToCareCovid19EntryBuilder.class)
```
jblefkove-va, line 21: I don't think you need this line. The special Lombok config in our shared mapper should use the builder to deserialize automatically.

facilities-collector/src/main/java/gov/va/api/lighthouse/facilitiescollector/AccessToCareCovid19Entry.java
```diff
@@ -0,0 +1,83 @@
...
+public class AccessToCareCovid19Entry {
+  @JsonProperty("FacilityLatitude")
+  BigDecimal latitude;
+
+  @JsonProperty("FacilityLongitude")
+  BigDecimal longitude;
```
jblefkove-va, line 27: Latitude and longitude are ununused, correct? Can we remove these? (They should fall back to the `@JsonAnySetter` method as a no-op.)

facilities-collector/src/main/java/gov/va/api/lighthouse/facilitiescollector/AccessToCareCovid19Entry.java
```diff
@@ -0,0 +1,83 @@
...
+
+  @JsonProperty("Facility")
+  String facility;
```
jblefkove-va, line 30: Since you are extracting `stationId` from `Facility` as part of the deserialization process, I don't think you still need to map this field. It should be sufficient to handle this in `@JsonAnySetter` as a third case when `equalsIgnoreCase(value, "facility")`

facilities-collector/src/main/java/gov/va/api/lighthouse/facilitiescollector/AccessToCareCovid19Entry.java
```diff
@@ -0,0 +1,83 @@
...
+
+  @JsonIgnore String stationId;
```
jblefkove-va, line 32: Why `@JsonIgnore`?

facilities-collector/src/main/java/gov/va/api/lighthouse/facilitiescollector/AccessToCareCovid19Entry.java
```diff
@@ -0,0 +1,83 @@
...
+
+  @JsonIgnore Integer confirmedCases;
+
+  @JsonIgnore Integer deaths;
+
+  /**
+   * Lombok doesnt play well with the JsonAnySetter annotation, so a custom builder needs to exist
+   * that deals with this.
+   */
+  @Slf4j
```
jblefkove-va, line 42: `@Slf4` on an inner class seems odd... should this go up on the parent?

facilities-collector/src/main/java/gov/va/api/lighthouse/facilitiescollector/AccessToCareCovid19Entry.java
```diff
@@ -0,0 +1,83 @@
...
+  @JsonPOJOBuilder(withPrefix = "")
```
jblefkove-va, line 43: I think this line is cruft now

facilities-collector/src/main/java/gov/va/api/lighthouse/facilitiescollector/AccessToCareCovid19Entry.java
```diff
@@ -0,0 +1,83 @@
...
+  public static class AccessToCareCovid19EntryBuilder {
+    /** Replaces lomboks build() method. */
+    public AccessToCareCovid19Entry build() {
+      return new AccessToCareCovid19Entry(
+          latitude,
+          longitude,
+          facility,
+          determineStationIdFromFacilityString(),
```
jblefkove-va, line 51: If you set `stationId` directly inside `saveUnknown()`, you should not need this custom `build()` method

facilities-collector/src/main/java/gov/va/api/lighthouse/facilitiescollector/AccessToCareCovid19Entry.java
```diff
@@ -0,0 +1,83 @@
...
+
+    /** Uses Jacksons JsonAnySetter annotation to map any unknown values to a map. */
+    @JsonAnySetter
+    public AccessToCareCovid19EntryBuilder saveUnknown(String key, Object value) {
+      // If value is null, theres no point in continuing
+      if (value == null) {
+        return this;
+      }
+      if (containsIgnoreCase(key, "confirmed")) {
+        confirmedCases = isNumeric(value.toString()) ? (Integer) value : null;
```
jblefkove-va, line 76: Use the other builder methods to populate these values. e.g. instead of `confirmedCases = ...`, do `return confirmedCases(...)`

facilities-collector/src/main/java/gov/va/api/lighthouse/facilitiescollector/AccessToCareCovid19Entry.java
```diff
@@ -0,0 +1,83 @@
...
+      String stnId = facility.substring(facility.indexOf("(") + 1, facility.indexOf(")"));
+      if (!stnId.matches("[A-Za-z0-9]+")) {
+        log.info("Collect access-to-care-covid19: Invalid station id ({}).", stnId);
+      }
+      return stnId;
+    }
+
+    /** Uses Jacksons JsonAnySetter annotation to map any unknown values to a map. */
+    @JsonAnySetter
+    public AccessToCareCovid19EntryBuilder saveUnknown(String key, Object value) {
```
jblefkove-va, line 70: Since each entry is still a flat key-value structure, the signature could be `saveUnknown(String key, String value)`. (That's not necessarily better than `Object value`... I don't know if it simplifies any of the logic.)

facilities-collector/src/main/java/gov/va/api/lighthouse/facilitiescollector/AccessToCareCovid19Entry.java
jhulbert-va: Working on these, I found something I completely overlooked. `JsonAnySetter` prevents unknown fields from causing a failure. My thought is to add some checks and throws into the builder. The expectation being that (at a minimum) facility is set and one of confirmed/deaths is set. Let me know if you disagree.

bschofield-va: That's not really a deserialization/builder responsibility. Validations are typically layered on using the Java Bean Validation framework.

jblefkove-va: Agreed, that is not how we're validating any of the other collector structures we deserialize. (e.g. any required field could be present, but explicitly set to `null`.) You can validate that individual entries make sense as you're populating the map of them in the HealthsCollector. (i.e. you can't put an entry in the map if it has no station ID)

facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/v0/Facility.java
```diff
@@ -131,6 +132,19 @@
     @Valid Address physical;
   }
 
+  @Data
+  @Builder
+  @JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY)
+  public static final class Covid19 {
+    @Schema(description = "Number of confirmed Covid-19 cases at a facility.", example = "46")
```
jblefkove-va, line 139: nitpick: COVID-19

facilities-collector/src/main/java/gov/va/api/lighthouse/facilitiescollector/AccessToCareCovid19Entry.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    /** Uses Jacksons JsonAnySetter annotation to map any unknown values to a map. */
+    @JsonAnySetter
+    public AccessToCareCovid19EntryBuilder saveUnknown(String key, String value) {
+
+      if (equalsIgnoreCase(key, "facility")) {
+        return stationId(determineStationIdFromFacilityString(value));
+      } else if (containsIgnoreCase(key, "confirmed")) {
+        return confirmedCases(isNumeric(value) ? Integer.parseInt(value) : null);
```
jblefkove-va, line 53: Super nitpick... `confirmedCases()` accepts `Integer`, while `parseInt(string)` returns primitive `int`. You can avoid the unnecessary autoboxing with `Integer.valueOf(string)`.

facilities-collector/src/main/java/gov/va/api/lighthouse/facilitiescollector/AccessToCareCovid19Entry.java
jhulbert-va: The caveat here is that the `CollectorHealthController` needs to be able to determine if anything has changed that'll break us. If all we do is validate the stationId, theres the very real chance that something changes and we can't map deaths/confirmedCases anymore, but don't notice it.

jhulbert-va: I could add validation logic to CollectorHealth.

jhulbert-va:
@bschofield-va @jblefkove-va I think its all or nothing for confirmedCases/deaths. My plan is to have the transformer deal with that.

facilities-collector/src/main/java/gov/va/api/lighthouse/facilitiescollector/AccessToCareCovid19Entry.java
```diff
@@ -0,0 +1,54 @@
...
+   */
+  public static class AccessToCareCovid19EntryBuilder {
+
+    /** Takes a facility of format "(station-id) facility-name" and pulls out the station_id. */
+    public String determineStationIdFromFacilityString(@NotNull String facility) {
+      return facility.substring(facility.indexOf("(") + 1, facility.indexOf(")"));
+    }
+
+    public Integer parseInt(String value) {
+      return isNumeric(value) ? Integer.valueOf(value) : null;
```
bschofield-va, line 37: I think I can break this.

bschofield-va:
```
jshell> String v = ""+((long) Integer.MAX_VALUE+1)
v ==> "2147483648"

jshell> org.apache.commons.lang3.StringUtils.isNumeric(v) ? Integer.parseInt(v): null
|  Exception java.lang.NumberFormatException: For input string: "2147483648"
|        at NumberFormatException.forInputString (NumberFormatException.java:68)
|        at Integer.parseInt (Integer.java:662)
|        at Integer.parseInt (Integer.java:776)
|        at (#4:1)
```

facilities-collector/src/main/java/gov/va/api/lighthouse/facilitiescollector/HealthsCollector.java
```diff
@@ -228,6 +232,39 @@ private ArcGisHealths loadArcGis() {
...
+    } catch (Exception e) {
+      log.info("Failed to read AccessToCare: Covid-19 data. {}", e.getMessage());
+      watch.stop();
+      return Optional.empty();
+    }
+
+    Map<String, AccessToCareCovid19Entry> map =
+        entries
+            .parallelStream()
+            .collect(Collectors.toMap(e -> upperCase("vha_" + e.stationId(), Locale.US), f -> f));
```
bschofield-va, line 259: `f -> f` is fine, but as FYI `identity()` does this too. See `java.util.function.Function.indentity()`

facilities-collector/src/main/java/gov/va/api/lighthouse/facilitiescollector/Transformers.java
```diff
@@ -26,6 +26,15 @@ static boolean allBlank(Object... values) {
     return true;
   }
 
+  static boolean anyBlank(Object... values) {
```
bschofield-va, line 29: `TransformersTest` could use a unit test for this

facilities-collector/src/test/java/gov/va/api/lighthouse/facilitiescollector/AccessToCareCovid19EntryTest.java
```diff
@@ -0,0 +1,142 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import com.fasterxml.jackson.databind.JsonMappingException;
+import gov.va.api.health.autoconfig.configuration.JacksonConfig;
+import lombok.SneakyThrows;
+import org.junit.Test;
+import org.junit.jupiter.api.Assertions;
+
+public class AccessToCareCovid19EntryTest {
```
bschofield-va, line 11: Can we see a few fuzzy death and confirmed attribute tests?

facilities-collector/src/main/java/gov/va/api/lighthouse/facilitiescollector/AccessToCareCovid19Entry.java
bschofield-va: I'll resolve this here, because I broken down below

bschofield-va:
ux chimed in on field descriptions:

https://lighthouseva.slack.com/archives/C010JJE7WDP/p1588102015076600?thread_ts=1587135570.069400&cid=C010JJE7WDP

facilities-collector/src/test/java/gov/va/api/lighthouse/facilitiescollector/AccessToCareCovid19EntryTest.java
jhulbert-va: If I'm understanding correctly, confirmed attribute tests are be in the `HealthTransformerTest`

bschofield-va:
# DO NOT MERGE
Until Covid 19 JSON data source is resolved

facilities-collector/src/main/java/gov/va/api/lighthouse/facilitiescollector/HealthsCollector.java
```diff
@@ -228,6 +235,38 @@ private ArcGisHealths loadArcGis() {
...
+    } catch (Exception e) {
+      log.info("Failed to read AccessToCare: Covid-19 data. {}", e.getMessage());
+      watch.stop();
+      return Optional.empty();
+    }
+    Map<String, AccessToCareCovid19Entry> map =
+        entries
+            .parallelStream()
+            .collect(
+                Collectors.toMap(e -> upperCase("vha_" + e.stationId(), Locale.US), identity()));
```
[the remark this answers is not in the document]
jblefkove-va: Note that for mental health numbers, if a duplicate key is detected while the map is being built up, this is logged as a warning. (And whichever one goes into the map last wins.)

facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/v0/Facility.java
```diff
@@ -131,6 +132,28 @@
...
   }
 
+  @Data
+  @Builder
+  @JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY)
+  public static final class Covid19 {
+    @Schema(
+        description =
+            "Number of COVID-19 cases at this facility. "
+                + "May include data from Veteran patients, non-Veteran patients, "
```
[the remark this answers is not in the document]
jhulbert-va: Looks like this needs to be updated with new info from Carrie anyway. Veteran is still capitalized though.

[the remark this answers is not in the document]
jhulbert-va: https://lighthouseva.slack.com/archives/C010JJE7WDP/p1588206142086400?thread_ts=1587135570.069400&cid=C010JJE7WDP

facilities-collector/src/main/java/gov/va/api/lighthouse/facilitiescollector/AccessToCareCovid19Entry.java
```diff
@@ -0,0 +1,79 @@
...
+@Value
+@Builder
+@AllArgsConstructor(access = AccessLevel.PRIVATE)
+@JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY)
+public class AccessToCareCovid19Entry {
+
+  /*
+   * Make the assumption stationId must be set for it to be a valid object.
+   */
+  @NonNull String stationId;
```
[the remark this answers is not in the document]
jhulbert-va: Sounds good. I'll add some custom logic to the Collector Healths endpoint as well then.

facilities-collector/src/main/java/gov/va/api/lighthouse/facilitiescollector/CollectorHealthController.java
```diff
@@ -153,20 +185,52 @@ public void clearCacheScheduler() {
...
+   */
+  private Health testAtcCovid19Health() {
+    HttpStatus statusCode;
+    String url =
+        UriComponentsBuilder.fromHttpUrl(atcCovidBaseUrl + "vacovid19summary.json").toUriString();
+    try {
+      ResponseEntity<String> response =
+          requestHealth(insecureRestTemplateProvider.restTemplate(), url);
+      statusCode = response.getStatusCode();
+      if (statusCode.value() == 200) {
```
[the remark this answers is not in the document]
jhulbert-va: Theoretically any status code could get here. The only time we'd skip it is if an error occurs (like we saw yesterday with the cert stuff).

facilities-collector/src/main/java/gov/va/api/lighthouse/facilitiescollector/CollectorHealthController.java
```diff
@@ -153,20 +185,52 @@ public void clearCacheScheduler() {
...
+
+  /**
+   * For the covid-19 response, the code tests not only its ability to reach out and get a 200
+   * response from the endpoint, but also that it is able to deserialize the response to a
+   * AccessToCareCovid19Entry. This will allow us to respond faster if the response changes.
+   */
+  private Health testAtcCovid19Health() {
+    HttpStatus statusCode;
+    String url =
+        UriComponentsBuilder.fromHttpUrl(atcCovidBaseUrl + "vacovid19summary.json").toUriString();
```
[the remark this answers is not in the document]
jhulbert-va: There are two files in there. Index json is automatically created by the terraform script and has basic bucket information. The actual file we'll be using in the bucket is vacovid19summary.json though.

facilities/src/main/java/gov/va/api/lighthouse/facilities/JacksonSerializersV0.java
```diff
@@ -210,6 +211,26 @@ public void serialize(
     }
   }
 
+  private static final class FacilityCovid19Serializer extends StdSerializer<Facility.Covid19> {
```
jblefkove-va, line 214: We don't need a custom serializer for the `Covid19` class itself. (The default behavior is fine. These are for backwards compatibility with the previous API, where the order of fields or presence of explicit null gets complicated.) Lines 276/521 are all we need to ensure this data gets written out.

PR #86, opened 2020-04-29
SHANK-000-add-logging-around-exceptions

jhulbert-va:


jhulbert-va:
There's not really a lot of other places logging can be added.

PR #87, opened 2020-04-29
Shank 000 insecure template for checks

jhulbert-va:


facilities-collector/src/main/java/gov/va/api/lighthouse/facilitiescollector/CollectorHealthController.java
```diff
@@ -89,36 +106,35 @@ public void clearCacheScheduler() {
     restTemplate.setRequestFactory(new HttpComponentsClientHttpRequestFactory());
     List<Health> downstreamServices =
         List.of(
-                Pair.of(
+                basicHealthCheck(
```
jblefkove-va, line 109: To be consistent with `HealthsCollector`, use the insecure rest template for the AccessToCare calls as well.

facilities-collector/src/main/java/gov/va/api/lighthouse/facilitiescollector/CollectorHealthController.java
```diff
@@ -137,7 +153,11 @@ public void clearCacheScheduler() {
     return ResponseEntity.ok(health);
   }
 
-  private Health testDownstreamHealth(String name, String url) {
+  private Supplier<Health> insecureHealthCheck(String name, String url) {
+    return () -> testDownstreamHealth(insecureRestTemplateProvider.restTemplate(), name, url);
+  }
+
+  private Health testDownstreamHealth(RestTemplate restTemplate, String name, String url) {
```
jblefkove-va, line 160: The first parameter is shadowing the `restTemplate` field on this class. Rename the parameter and/or make this method static.

facilities-collector/src/main/java/gov/va/api/lighthouse/facilitiescollector/CollectorHealthController.java
```diff
@@ -89,36 +106,35 @@ public void clearCacheScheduler() {
     restTemplate.setRequestFactory(new HttpComponentsClientHttpRequestFactory());
```
jblefkove-va, line 106: I'm not 100% clear on what this line is for, but... should we do the same for the insecure rest template as well?

facilities-collector/src/main/java/gov/va/api/lighthouse/facilitiescollector/CollectorHealthController.java
```diff
@@ -89,36 +106,35 @@ public void clearCacheScheduler() {
...
     List<Health> downstreamServices =
         List.of(
-                Pair.of(
```
jblefkove-va, line 92: I think the implementation with the `Pair`s is easier to follow than `Supplier`s. Can we create something like a `HealthCheck` class in here that has three fields (`restTemplate`, `name`, `url`) and use that?

bschofield-va: ftw ... I don't really find more or less easy to follow. I like the supplier because it captures the _idea_ of doing a task (get health status) vs. the pair that captures the _details_ required to execute a specific task (here's my name and url, get the status in the one and only one way you know how to).

I don't feel strongly, so if Peter does then use the pair otherwise 🤷

jblefkove-va: It's already switched to supplier in the COVID-19 PR, so that's fine

PR #89, opened 2020-04-30
SHANK-587 default content type

jblefkove-va:
https://vasdvp.atlassian.net/browse/SHANK-587


facilities-tests/src/test/java/gov/va/api/lighthouse/facilities/tests/FacilitiesIT.java
```diff
@@ -26,20 +27,15 @@
...
-  @Test
-  @Category({AllFacilities.class})
-  public void allAsCsv() {
-    final String request = "v0/facilities/all";
-    makeRequest("text/csv", request, 200);
-  }
-
   @Test
   @Category({AllFacilities.class})
   public void allAsJson() {
```
jhulbert-va, line 32: This test name should change if we add `text/csv` to it.

PR #91, opened 2020-05-12
Api 839 remove covid

jhulbert-va:
https://vajira.max.gov/browse/API-839

Not seeing anything covid related when testing the changes locally.


jblefkove-va:
Remove it from the `README`s as well?

jhulbert-va:
> Remove it from the `README`s as well?

I'm tempted to say no. The only thing we're doing here is removing it from documentation and the output. All processes outlined in the readme still occur behind the scenes.

PR #93, opened 2020-05-19
Api 901 omit office cemeteries

jhulbert-va:
https://vajira.max.gov/browse/API-901


facilities-collector/src/main/java/gov/va/api/lighthouse/facilitiescollector/CemeteriesCollector.java
```diff
@@ -30,6 +32,7 @@
   @SneakyThrows
   public Collection<Facility> collect() {
     return requestArcGisCemeteries().features().stream()
+        .filter(c -> !equalsIgnoreCase(c.attributes().siteType(), "office"))
```
jblefkove-va, line 35: one line ftw

bschofield-va: streams are sgood

PR #98, opened 2020-05-26
Api 948 better collector error messages

eclendenning-va:


facilities-collector/src/main/java/gov/va/api/lighthouse/facilitiescollector/HealthsCollector.java
```diff
@@ -118,7 +118,7 @@ static void putStopCode(ResultSet resultSet, ListMultimap<String, StopCode> map)
             .build());
   }
 
-  Collection<Facility> healths() {
+  Collection<Facility> healths() throws HealthsCollectorException {
     ListMultimap<String, AccessToCareEntry> accessToCareEntries = loadAccessToCare();
```
jhulbert-va, line 122: Why not leave sneakyThrows on the load  methods and catch/throw the `HealthCollectorException` here.

facilities-collector/src/main/java/gov/va/api/lighthouse/facilitiescollector/StateCemeteriesCollector.java
```diff
@@ -82,4 +85,10 @@ public String deserialize(JsonParser p, DeserializationContext ctxt) {
           });
     }
   }
+
+  static final class StateCemeteriesException extends RuntimeException {
```
jblefkove-va, line 89: `StateCemeteriesCollectorException` to be consistent with the others

facilities-collector/src/main/java/gov/va/api/lighthouse/facilitiescollector/HealthsCollector.java
jblefkove-va: SneakyThrows ftw, checked exceptions ftl

jblefkove-va:
There is a subtlety here... in the main `CollectController.collectFacilities()` method, before each of the individual collectors is invoked, there is the setup of the `loadWebsites()` and `loadVast()` methods. If either of those throws an exception, we want to catch that and rewrap it as well.

facilities-collector/src/main/java/gov/va/api/lighthouse/facilitiescollector/BenefitsCollector.java
```diff
@@ -59,19 +56,29 @@ private ArcGisBenefits requestArcGisBenefits() {
...
+          benefits.features().size());
+      return benefits;
+    } catch (Exception e) {
+      throw new BenefitsCollectorException(e.getMessage());
+    }
+  }
+
+  static final class BenefitsCollectorException extends RuntimeException {
+    public BenefitsCollectorException(String errorMessage) {
+      super(errorMessage);
```
jblefkove-va, line 81: In general, rewrap the whole exception. This will preserve the complete stack trace in the logs, not just the underlying message:

```
class BenefitsCollectorException extends RuntimeException {
    public BenefitsCollectorException(Throwable cause) {
      super(cause);
    }
  }
```

jblefkove-va:
@jhulbert-va 100% agree on points 2 and 3 above. (And standardizing on a single `collect` method name is low-impact, so I'm okay lumping that into this PR.) 

For point 1, I do think exception subclassing is a less error-prone way of providing context about which part of collecting failed. However, for better organization, I would group all of the exceptions (including preexisting `CdwException`) into a utility class, similar to `ResourceExceptions` in data-query. 

Also, in a previous comment, I mention that `CollectController.collectFacilities()` itself may generate exceptions outside of the individual collector classes. So there should be another exception class for that case, but a good name doesn't jump out at me. I'm thinking there should be a generic `CollectorException` class that `BenefitsCollectorException`, etc. extend, and if there's a problem, say, parsing `websites.csv`, just throw `CollectorException` itself.

facilities-collector/src/main/java/gov/va/api/lighthouse/facilitiescollector/CollectController.java
```diff
@@ -86,26 +86,30 @@ public CollectController(
...
-      Iterable<CSVRecord> rows = CSVFormat.DEFAULT.withFirstRecordAsHeader().parse(reader);
-      Map<String, String> map = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
-      for (CSVRecord row : rows) {
-        String id = trimToNull(row.get(WebsiteCsvHeaders.id));
-        String url = trimToNull(row.get(WebsiteCsvHeaders.url));
-        checkState(id != null, "Website %s missing ID", url);
-        checkState(url != null, "Website %s missing url", id);
-        checkState(!map.containsKey(id), "Website %s duplicate", id);
-        map.put(id, url);
+    try {
```
jhulbert-va, line 89: I think this try catch should be in `collectFacilities()`

facilities-collector/src/main/java/gov/va/api/lighthouse/facilitiescollector/CollectController.java
```diff
@@ -177,13 +181,17 @@ public CollectorFacilitiesResponse collectFacilities() {
...
 
   private List<VastEntity> loadVast() {
-    final Stopwatch watch = Stopwatch.createStarted();
-    ImmutableList<VastEntity> entities = ImmutableList.copyOf(vastRepository.findAll());
-    log.info(
-        "Loading vast took {} millis for {} entries",
-        watch.stop().elapsed(TimeUnit.MILLISECONDS),
-        entities.size());
-    return entities;
+    try {
```
jhulbert-va, line 184: Same here.

facilities-collector/src/main/java/gov/va/api/lighthouse/facilitiescollector/BenefitsCollector.java
```diff
@@ -26,17 +26,20 @@
...
-    return requestArcGisBenefits().features().stream()
-        .map(
-            facility ->
-                BenefitsTransformer.builder()
-                    .arcgisFacility(facility)
-                    .csvWebsite(websites.get("vba_" + facility.attributes().facilityNumber()))
-                    .build()
-                    .toFacility())
-        .collect(Collectors.toList());
+  public Collection<Facility> collect() throws CollectorExceptions.BenefitsCollectorException {
```
jblefkove-va, line 29: SneakyThrows!

facilities-collector/src/main/java/gov/va/api/lighthouse/facilitiescollector/CollectController.java
jblefkove-va, line 89: This would be easier to read as one `try` block wrapping the two lines `loadWebsites()` and `loadVast()` in the top-level `collectFacilities()` method

facilities-collector/src/test/java/gov/va/api/lighthouse/facilitiescollector/HealthsCollectorTest.java
```diff
@@ -296,6 +293,52 @@ public void collect() {
                     .build()));
   }
 
+  @Test(expected = CollectorExceptions.HealthsCollectorException.class)
+  public void exception() {
+    RestTemplate insecureRestTemplate = mock(RestTemplate.class);
```
jhulbert-va, line 298: why not use this mock and throw a `RestClientException` when it requests access to care?

PR #99, opened 2020-05-28
collect vet centers from vast instead of ArcGis

eclendenning-va:
https://vajira.max.gov/browse/API-238


facilities-collector/src/main/java/gov/va/api/lighthouse/facilitiescollector/VetCentersCollector.java
```diff
@@ -1,76 +1,27 @@
...
-            .exchange(url, HttpMethod.GET, new HttpEntity<>(new HttpHeaders()), String.class)
-            .getBody();
-    ArcGisVetCenters vetCenters =
-        JacksonConfig.createMapper().readValue(response, ArcGisVetCenters.class);
-    log.info(
-        "Loading vet centers took {} millis for {} features",
-        totalWatch.stop().elapsed(TimeUnit.MILLISECONDS),
-        vetCenters.features().size());
-    return vetCenters;
+    return vastEntities.stream()
```
jblefkove-va, line 18: This method could still throw an exception, e.g. if there's a bug in the transformer. So we should keep the `try/catch` and `VetCentersCollectorException`

jblefkove-va:
Update `facilities-mock-services/real-dependency-service-requests`

PR #100, opened 2020-05-29
API-991 Additional fields for vet centers

jblefkove-va:
https://vajira.max.gov/browse/API-991


facilities-collector/src/main/java/gov/va/api/lighthouse/facilitiescollector/VetCenterTransformer.java
```diff
@@ -17,23 +18,43 @@
...
+        .state(upperCase(vast.state(), Locale.US))
+        .address1(vast.address2())
+        .address2(vast.address3())
         .build();
   }
 
+  private String addressZip() {
+    String zip = vast.zip();
+    String zipPlus4 = vast.zip4();
+    if (isNotBlank(zip) && isNotBlank(zipPlus4) && !zipPlus4.matches("^[0]+$")) {
```
taylorrassmannlibertyits, line 52: Is there any regex needed on the "zip" also?

jblefkove-va: There could be, but we have not encountered that. We have seen missing values for `zip4` populated as zeroes, though.

PR #106, opened 2020-06-24
Api 1324 facilities backend health controller

jhulbert-va:
[API-1324](https://vajira.max.gov/browse/API-1324)


facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilitiesHealthController.java
```diff
@@ -0,0 +1,87 @@
...
+  @CacheEvict(value = "health")
+  public void clearCacheScheduler() {
+    if (hasCachedRecently.getAndSet(false)) {
+      log.info("Clearing downstream health cache.");
+    }
+  }
+
+  private Health facilitiesCollectionHealth() {
+    Instant lastSuccessfulUpdate = repository.findLastUpdated();
+    Instant outOfDate = Instant.now().minus(24, ChronoUnit.HOURS);
```
jblefkove-va, line 57: `outOfDate` seems more like a boolean name. How about something like `oldestValidDate`?

facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilitiesHealthController.java
```diff
@@ -0,0 +1,87 @@
...
+    return buildHealth(
+        "Facilities Collection",
+        lastSuccessfulUpdate != null && lastSuccessfulUpdate.isAfter(outOfDate),
```
jblefkove-va, line 60: For awareness, once this deploys to prod/lab, the health check will fail until we force another collection. We will want to do that before setting up the corresponding Pingdom check.

facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilitiesHealthController.java
```diff
@@ -0,0 +1,87 @@
...
+   * The entry point for the FacilitiesHealthController.
+   *
+   * @return A list of healthy/unhealthy downstream services (with matching statusCode).
+   */
+  @GetMapping("/health")
+  @Cacheable("health")
+  public ResponseEntity<Health> health() {
+    hasCachedRecently.set(true);
+    // If another health check is added, use parallelStream()
+    List<Health> downstreamHealth = List.of(facilitiesCollectionHealth());
```
jblefkove-va, line 74: YAGNI... let's remove the list and just interact with the single item directly.

facilities/src/main/resources/db/changelog/db.changelog-master.yaml
```diff
@@ -229,3 +229,17 @@ databaseChangeLog:
...
+      id: 7
+      author: shanktopus
+      changes:
+        addColumn:
+          tableName: facility
+          schemaName: app
+          columns:
+            - column:
+                name: last_updated
+                type: smalldatetime
```
jblefkove-va, line 243: `missing_timestamp` above is `bigint`

facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilityManagementController.java
```diff
@@ -57,6 +57,7 @@ static FacilityEntity populate(FacilityEntity record, Facility facility) {
     record.servicesFromServiceTypes(serviceTypesOf(facility));
     record.facility(FacilitiesJacksonConfig.createMapper().writeValueAsString(facility));
     record.missingTimestamp(null);
+    record.lastUpdated(Instant.now());
```
jblefkove-va, line 60: Would it make sense to call `Instant.now()` at a higher level and pass this in? It seems silly that we're going to invoke this 2400 times and populate the table with subtly different timestamps.

facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilitiesHealthController.java
jblefkove-va, line 60: Edge case... is collection considered healthy if `lastSuccessfulUpdate` is exactly 24 hours old?

facilities/src/test/java/gov/va/api/lighthouse/facilities/FacilitiesHealthControllerTest.java
```diff
@@ -0,0 +1,67 @@
...
+import lombok.Builder;
+import lombok.Value;
+import org.junit.jupiter.api.Test;
+import org.springframework.boot.actuate.health.Health;
+import org.springframework.boot.actuate.health.Status;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
+
+public class FacilitiesHealthControllerTest {
+  FacilityRepository repository = mock(FacilityRepository.class);
```
jblefkove-va, line 21: I know Bryan is not a fan of them, but I think this test should be `@DataJpaTest` with a real repository, so that the `@Query("select max(e.lastUpdated) from #{#entityName} e")` up in `FacilityRepository` gets exercised.

facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilitiesHealthController.java
jhulbert-va: I would say yes.

facilities/src/main/resources/db/changelog/db.changelog-master.yaml
jhulbert-va: I can make this a bigint, smalldatetime is easier to work with though and thats the same type that VAST uses.

facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilitiesHealthController.java
jblefkove-va: I agree

facilities/src/main/resources/db/changelog/db.changelog-master.yaml
jblefkove-va: I thought dates as longs was the general preference, but I don't have strong feelings on it. We do have to interact with some database schemas (that we don't own) that use date schemas... mapping to `Instant` in the entity is appropriate there.

facilities/src/test/java/gov/va/api/lighthouse/facilities/FacilitiesHealthControllerTest.java
bschofield-va: Both statements are true! In _most_ cases, I don't think we need `DataJpaTests`, but since this now has explicit DB logic (in the form of a query) it should be exercised _somewhere_. If that means this test stays _as is_ and we add another test to specifically exercise this bit of repository (`@DataJpaTest FacilityRepositoryIT`).

facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilitiesHealthController.java
```diff
@@ -0,0 +1,87 @@
...
+        lastSuccessfulUpdate != null && lastSuccessfulUpdate.isAfter(outOfDate),
+        Map.of("status", "lastUpdated: " + lastSuccessfulUpdate));
+  }
+
+  /**
+   * The entry point for the FacilitiesHealthController.
+   *
+   * @return A list of healthy/unhealthy downstream services (with matching statusCode).
+   */
+  @GetMapping("/health")
```
bschofield-va, line 69: I'm not sure I like this path ..
`/v0/facilities/health`
It seems to imply the health of the Facilities API, which is not true.

Collector uses `/services/va_facilities/collector/health`
Should this check be `/services/va_facilities/collector/status` or `collection/status` or `collection/health`

facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilitiesHealthController.java
```diff
@@ -0,0 +1,87 @@
...
+  @Cacheable("health")
+  public ResponseEntity<Health> health() {
+    hasCachedRecently.set(true);
+    // If another health check is added, use parallelStream()
+    List<Health> downstreamHealth = List.of(facilitiesCollectionHealth());
+    Health overallHealth =
+        buildHealth(
+            "Downstream Services",
```
bschofield-va, line 76: This isn't a downstream service. I think the structure of this should not match the collector's downstream structure. That is, there isn't an _overall_ health (at least not now)

facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilitiesHealthController.java
jhulbert-va: Sticking with the convention we've used for the other health controllers, I would vote for `/collection/health`

facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilitiesHealthController.java
bschofield-va: I think the output is just 
```
      {
        "description": "Facilities Collection",
        "status": "DOWN",
        "details": {
          "name": "Facilities Collection",
          "lastUpdated":  "2020-06-22T09:00:41.412165Z",
          "time": "2020-06-24T19:20:41.412165Z"
        }
 ```

File: facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilitiesHealthController.java [hunk 9]
```diff
@@ -0,0 +1,86 @@
...
+    }
+  }
+
+  private Health facilitiesCollectionHealth() {
+    Instant lastSuccessfulUpdate = repository.findLastUpdated();
+    Instant oldestValidDate = Instant.now().minus(24, ChronoUnit.HOURS);
+    return buildHealth(
+        "Facilities Collection",
+        lastSuccessfulUpdate != null && lastSuccessfulUpdate.isAfter(oldestValidDate),
+        Map.of("status", "lastUpdated: " + lastSuccessfulUpdate));
```
bschofield-va, line 61: This produces: `"status": "lastUpdated: <time>"` which is weird. Should just be `"lastUpdated" : "<time>"`

jhulbert-va:
```
$ curl http://localhost:8085/collection/status -s | jq .
{
  "description": "Facilities Collection",
  "status": "UP",
  "details": {
    "name": "Facilities Collection",
    "status": "lastUpdated: 2020-06-24T19:11:00Z",
    "time": "2020-06-24T20:23:02.352900Z"
  }
}
```

facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilitiesHealthController.java
```diff
@@ -0,0 +1,80 @@
...
+            : HttpStatus.SERVICE_UNAVAILABLE;
+    return ResponseEntity.status(healthStatus).body(collectionStatus);
+  }
+
+  private Health facilitiesCollectionStatus() {
+    Instant lastSuccessfulUpdate = repository.findLastUpdated();
+    Instant oldestValidDate = Instant.now().minus(24, ChronoUnit.HOURS);
+    return buildHealth(
+        "Facilities Collection",
+        lastSuccessfulUpdate != null && lastSuccessfulUpdate.isAfter(oldestValidDate),
```
jblefkove-va, line 77: The edge case I mentioned before is still here. This will be down if `lastSuccessfulUpdate.toEpochMilli() == Instant.now().minus(24, ChronoUnit.HOURS).toEpochMilli()`

facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilitiesHealthController.java
```diff
@@ -0,0 +1,80 @@
...
+    HttpStatus healthStatus =
+        collectionStatus.getStatus().equals(Status.UP)
+            ? HttpStatus.OK
+            : HttpStatus.SERVICE_UNAVAILABLE;
+    return ResponseEntity.status(healthStatus).body(collectionStatus);
+  }
+
+  private Health facilitiesCollectionStatus() {
+    Instant lastSuccessfulUpdate = repository.findLastUpdated();
+    Instant oldestValidDate = Instant.now().minus(24, ChronoUnit.HOURS).minusSeconds(1);
```
jblefkove-va, line 74: Minus-one-second seems hacky. How about this:
```
Instant oldestValidDate = Instant.now().minus(24, ChronoUnit.HOURS);
boolean outOfDate = lastSuccessfulUpdate == null || oldestValidDate.isAfter(lastSuccessfulUpdate);
return buildHealth("Facilities Collection", !outOfDate, ...);
```

PR #110, opened 2020-06-30
API-1398 CDW debug endpoint for stop codes table

jblefkove-va:
https://vajira.max.gov/browse/API-273


bschofield-va:
should there be a unit test for this?

jblefkove-va:
This new method will replace the existing one in a subsequent PR. The tests for the existing one can be updated for this.

PR #111, opened 2020-07-02
API-1399 stop codes ETL table

jblefkove-va:
https://vajira.max.gov/browse/API-273


facilities-collector/src/test/java/gov/va/api/lighthouse/facilitiescollector/VastRepositoryTest.java
```diff
@@ -18,7 +18,7 @@ public void lastUpdated() {
     var aLongTimeAgo = Instant.parse("2020-01-20T02:20:00Z");
     repository.save(vastEntity(1L, aLongTimeAgo));
     assertThat(repository.findLastUpdated()).isEqualTo(aLongTimeAgo);
-    var now = Instant.now();
```
ilaflamme-va, line 21: oops.

jblefkove-va: Oddly this test doesn't pass on my machine. The last line winds up comparing two instants with different nanoseconds. Probably a windows thing. Creating an Instant at current time with only millisecond granularity works around it.

PR #112, opened 2020-07-08
Api 1392 fix logging threads facilities collector

taylorrassmannlibertyits:


jblefkove-va:
Let's add a test for the health endpoint to `CollectorIT`

PR #117, opened 2020-07-21
Api 1575 remove covid 19 references

taylorrassmannlibertyits:
https://vajira.max.gov/browse/API-1575


jblefkove-va:
Remove SCRAPER from system-components diagram:
image

PR #119, opened 2020-07-27
F: @NotNull vs @NonNull

taylorrassmannlibertyits:
https://vajira.max.gov/browse/API-1671
Api 1671 change notnull to nonnull


jblefkove-va:
BenefitsTransformer and CemeteriesTransformer can also be updated

taylorrassmannlibertyits:
> BenefitsTransformer and CemeteriesTransformer can also be updated

Will be updated shortly thanks!

PR #121, opened 2020-07-29
Update GH Diagram

taylorrassmannlibertyits:
https://vajira.max.gov/browse/API-1641
Updated UML diagrams with better descriptions.


jblefkove-va:
Change "Cemeteries and benefit centers" label to "National cemeteries and benefit centers"

jblefkove-va:
While we're in here... the label `kubernetes cron` for the facilities reload timer is out of date. Change that to `AWS Lambda cron`.

PR #126, opened 2020-07-31
API-1672 Increase timeout for starting apps for facilities-tests

jnguy-va:
https://vajira.max.gov/browse/API-1672


facilities-tests/pom.xml
```diff
@@ -96,6 +96,7 @@
               <skip>${sentinel.skipLaunch}</skip>
               <processLogFile>${project.build.directory}/mock-services.log</processLogFile>
               <waitForInterrupt>false</waitForInterrupt>
+              <waitAfterLaunch>60</waitAfterLaunch>
```
jblefkove-va, line 99: This is for `mock-services`, which starts fairly quickly even on GFE. The slow one is `facilities-process` further down.

PR #137, opened 2020-08-05
API-1690 Refactor collector app into facilities app

jblefkove-va:
https://vajira.max.gov/browse/API-1690

Must merge with DU changes: https://github.com/department-of-veterans-affairs/lighthouse-facilities-deployment/pull/45/

Deployer branch: https://github.com/department-of-veterans-affairs/health-apis-deployer/compare/API-1811-remove-facilities-collector-2


facilities-tests/src/test/java/gov/va/api/lighthouse/facilities/tests/CollectorIT.java
```diff
@@ -14,42 +15,40 @@
...
-    makeRequest("collector/health", 200);
+    var items = makeRequest("internal/collector/facilities", 200).expectListOf(Facility.class);
+    assertThat(items.size()).isGreaterThan(2000);
   }
 
   @Test
   void mentalHealth() {
-    var items = makeRequest("mental-health-contact", 200).expectListOf(Map.class);
+    var items =
+        makeRequest("internal/collector/mental-health-contact", 200).expectListOf(Map.class);
```
jblefkove-va, line 39: The collector CDW endpoints (mental-health-contact, stop-code, and vast) have moved to `/internal/collector`. These are only for our internal debugging, so nothing else needs to be updated.

facilities-tests/src/test/java/gov/va/api/lighthouse/facilities/tests/CollectorIT.java
```diff
@@ -14,42 +15,40 @@
...
 
   @Test
   void collectFacilities() {
-    makeRequest("collect/facilities", 200).expectValid(CollectorFacilitiesResponse.class);
-  }
-
-  @Test
-  void collectHealth() {
-    makeRequest("collector/health", 200);
+    var items = makeRequest("internal/collector/facilities", 200).expectListOf(Facility.class);
```
jblefkove-va, line 32: The facilities collect endpoint, `/collect/facilities/`, is now `internal/collector/facilities`. Since the facilities app is no longer invoking it, it's just for our internal debugging as well.

facilities/src/main/java/gov/va/api/lighthouse/facilities/HealthController.java
```diff
@@ -0,0 +1,237 @@
...
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RestController;
+import org.springframework.web.client.ResourceAccessException;
+import org.springframework.web.client.RestTemplate;
+import org.springframework.web.util.UriComponentsBuilder;
+
+@Slf4j
+@RestController
+@RequestMapping(produces = "application/json")
+public class HealthController {
```
jblefkove-va, line 36: This is a combination of the two backend health checks - what was previously `CollectorHealthController` and `FacilitiesHealthController`.

facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/DbConfig.java
```diff
@@ -0,0 +1,25 @@
...
+import org.springframework.jdbc.core.JdbcTemplate;
+
+@Configuration
+class DbConfig {
+  @Bean
+  JdbcTemplate jdbcTemplate(
+      @Value("${spring.fc-datasource.driver-class-name}") String driverClassName,
+      @Value("${spring.fc-datasource.password}") String password,
+      @Value("${spring.fc-datasource.url}") String url,
+      @Value("${spring.fc-datasource.username}") String username) {
```
jblefkove-va, line 16: Configuring one Spring app to use two different databases is tricky. My solution is to use a `JdbcTemplate` for CDW, with manual configuration of its data source; this avoid conflict with the data source that Spring automatically configures for the facilities operation database. This avoids changing the facilities operational database configuration, or adding additional configuration there.

facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/FacilitiesCollector.java
```diff
@@ -27,30 +28,18 @@
...
- * easy parsing.
- */
 @Slf4j
-@Validated
-@RestController
-@SuppressWarnings("WeakerAccess")
-@RequestMapping(value = "/collect", produces = "application/json")
-public class CollectController {
+@Component
+public class FacilitiesCollector {
```
jblefkove-va, line 36: Most of the logic of this class is unchanged. It was the main controller in the collector app; now it is a component that gets injected where needed in the facilities app.

facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/FacilitiesCollector.java
```diff
@@ -168,26 +198,56 @@ public CollectorFacilitiesResponse collectFacilities() {
...
-                .collect(toList()))
-        .build();
+    return Streams.stream(Iterables.concat(benefits, cemeteries, healths, stateCems, vetCenters))
+        .sorted((left, right) -> left.id().compareToIgnoreCase(right.id()))
+        .collect(toList());
   }
 
   private List<VastEntity> loadVast() {
     Stopwatch watch = Stopwatch.createStarted();
-    List<VastEntity> entities = ImmutableList.copyOf(vastRepository.findAll());
```
jblefkove-va, line 181: Unfortunately `VastRepository` is no longer available since we have to use `JdbcTemplate` to access CDW. So some boilerplate JDBC code was required here.

facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/VastEntity.java
```diff
@@ -2,46 +2,25 @@
...
 @Builder
-@Table(name = "vast", schema = "app")
-@NoArgsConstructor(access = AccessLevel.PUBLIC)
-@EqualsAndHashCode(onlyExplicitlyIncluded = true)
-@AllArgsConstructor(access = AccessLevel.PRIVATE)
-public class VastEntity {
-  @Id @EqualsAndHashCode.Include private Long vastId;
-
-  @Column(name = "VCTR2")
+final class VastEntity {
```
jblefkove-va, line 11: this class is no longer a JPA entity, just a holder for all the results of the JDBC call

facilities/src/main/java/gov/va/api/lighthouse/facilities/HomeController.java
```diff
@@ -43,6 +52,21 @@ public Metadata metadata() {
         .build();
   }
 
+  @SneakyThrows
+  private String openapiContent() {
+    try (InputStream is = openapi.getInputStream()) {
```
taylorrassmannlibertyits, line 57: What happens if there is an exception here? Should there be a throws of some kind?

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalCollectorController.java
```diff
@@ -42,38 +50,41 @@
...
-      throw new CollectorExceptions.CdwException(ex);
+      throw new CdwException(ex);
     }
   }
 
-  /** Stop codes for debugging. */
-  @SneakyThrows
-  @RequestMapping(value = "/stop-code", produces = "application/json", method = RequestMethod.GET)
-  public List<Map<String, String>> stopCodes() {
+  @GetMapping(value = "/stop-code")
```
taylorrassmannlibertyits, line 67: This doesn't need the `internal/...` path due ? Is this because of how the mapping works?

jblefkove-va: The `/internal/collector/` path is specified up on the controller class, so it applies to all the request methods in here.

facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/FacilitiesCollector.java
```diff
@@ -168,26 +198,56 @@ public CollectorFacilitiesResponse collectFacilities() {
...
+    List<VastEntity> entities =
+        ImmutableList.copyOf(
+            jdbcTemplate.query(
```
taylorrassmannlibertyits, line 210: In the future for JDBC code, maybe we have a support class for different boilerplant usages.

facilities/src/main/java/gov/va/api/lighthouse/facilities/HomeController.java
jblefkove-va: The try-with-resources means we guarantee to close the stream when we're done with it. If an exception did happen here, it would escape and be mapped to a 500 error response for retrieving the Swagger.

In practice I don't think this would happen; if there is a problem with the swagger file missing or being in the wrong place, the app will simply fail to start up at all because of the mapped resource in the constructor. (You may have seen this while trying to launch the app from your IDE; if you do a clean and rebuild in the IDE using its compiler, not Maven, the swagger does not get regenerated, and the app will fail to start with an error indicating that `openapi.json` could not be found.)

facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/FacilitiesCollector.java
jblefkove-va: Yes, we can revisit that; there are two stories in the backlog to start getting national cemeteries and benefits centers from CDW instead of ArcGIS endpoints.

PR #138, opened 2020-08-06
Add PSSG request/updater to facilities test

taylorrassmannlibertyits:
https://vajira.max.gov/browse/API-1815
Added PssgRequestBot to facilities-test
Updated PSSG bands json test files to reflect new format.


jblefkove-va:
mock services `pssg-drive-time-bands-2.json` should be updated as well

facilities-tests/src/test/java/gov/va/api/lighthouse/facilities/tests/PssgRequestBot.java
```diff
@@ -0,0 +1,91 @@
...
+    final int RECORD_COUNT = 25;
+    PssgBandFeatures response;
+
+    do {
+      String requestString =
+          "arcgis2/rest/services/Portal/MonthlyVAST_TTB/FeatureServer/0/"
+              + "query?f=json&inSR=4326&outSR=4326&orderByFields=Name&outFields=*&resultOffset="
+              + offset
+              + "&returnCountOnly=false&returnDistinctValues=false&"
+              + "returnGeometry=true&where=1=1&resultRecordCount=25";
```
jblefkove-va, line 41: do you want to reuse the RECORD_COUNT var here?

facilities-tests/src/test/java/gov/va/api/lighthouse/facilities/tests/PssgRequestBot.java
```diff
@@ -0,0 +1,91 @@
...
+public class PssgRequestBot {
+  private static final ObjectMapper MAPPER = JacksonConfig.createMapper();
+
+  @SneakyThrows
+  public static void main(String[] args) {
+
+    List<PssgDriveTimeBand> bands = new ArrayList<>();
+    final String BASE_URI = "https://vaww.pssg.med.va.gov/";
+    int offset = 0;
+    final int RECORD_COUNT = 25;
```
jblefkove-va, line 32: Change the mock-services batch size from 30 to 25

facilities-tests/src/test/java/gov/va/api/lighthouse/facilities/tests/PssgRequestBot.java
```diff
@@ -0,0 +1,91 @@
...
+
+      log.info("Offset: " + offset);
+
+    } while (!response.features.isEmpty());
+
+    log.info("About to send massive post!");
+
+    SystemDefinitions.Service svc = SystemDefinitions.systemDefinition().facilitiesManagement();
+
+    ExpectedResponse.of(
```
jblefkove-va, line 68: Move this up into the loop. You want to post each small batch of bands after retrieving it; trying to hold all the bands in memory at the same time is going to be an OutOfMemoryException.

facilities-tests/src/test/java/gov/va/api/lighthouse/facilities/tests/PssgRequestBot.java
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  /** The response actually contains more fields, but we don't use them and have not been mapped. */
+  @Data
+  @Builder
+  @JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY)
+  @JsonIgnoreProperties(ignoreUnknown = true)
+  @AllArgsConstructor
+  @NoArgsConstructor
+  private static class PssgBandFeatures {
```
jblefkove-va, line 88: I think the only annotations you need here are `Data` and `JsonIgnoreProperties`. (You're not going to be building these manually, `JsonAutoDetect` is to help with Swagger, and all-args/no-args constructors are implicit in `Data`.)

taylorrassmannlibertyits:
> pssg-drive-time-bands-2

That one is an empty json array. I think it is used to test that data scenario?

facilities-tests/src/test/java/gov/va/api/lighthouse/facilities/tests/PssgRequestBot.java
taylorrassmannlibertyits: Yes we do!

facilities-tests/src/test/java/gov/va/api/lighthouse/facilities/tests/PssgRequestBot.java
taylorrassmannlibertyits: I think this is already 25?

facilities-tests/src/test/java/gov/va/api/lighthouse/facilities/tests/PssgRequestBot.java
taylorrassmannlibertyits: Ah ok, that's what I was thinking originally. I didn't communicate it well enough. Will do!

facilities-tests/src/test/java/gov/va/api/lighthouse/facilities/tests/PssgRequestBot.java
taylorrassmannlibertyits: Sounds good will update.

facilities-tests/src/test/java/gov/va/api/lighthouse/facilities/tests/PssgRequestBot.java
jblefkove-va: It's 25 here, but it is 30 in the hard-coded urls in mock-services

taylorrassmannlibertyits: Ah I see now. Will update.

jblefkove-va:
> > pssg-drive-time-bands-2
> 
> That one is an empty json array. I think it is used to test that data scenario?

It is empty, it is used to facilitate our use case of querying PSSG for batches of bands until we get an empty response back, which indicates we are done. Since the response format changed, it is no longer an empty array, `[ ]`, but a response like this:

```
{
  ...
  headers
  ...
  features: []
}
```

You can force PSSG to give you an empty response by querying it with a very high `resultOffset`

facilities-mock-services/real-dependency-service-requests
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
```
jblefkove-va, line 1: this file must have been added back by mistake during the merge... delete this

taylorrassmannlibertyits: Removed!

facilities-tests/src/test/java/gov/va/api/lighthouse/facilities/tests/PssgRequestBot.java
```diff
@@ -0,0 +1,81 @@
...
+        log.info("About to send POST!");
+
+        ExpectedResponse.of(
+                RestAssured.given()
+                    .baseUri(svc.url())
+                    .relaxedHTTPSValidation()
+                    .header("client-key", System.getProperty("client-key", CLIENT_KEY_DEFAULT))
+                    .contentType("application/json")
+                    .body(MAPPER.writeValueAsString(response.features))
+                    .request(Method.POST, "internal/management/bands"))
```
jblefkove-va, line 60: I realized what was causing the 503... this needs to be `.request(Method.POST, svc.urlWithApiPath() + "internal/management/bands"))`. It works as is locally, because the internal api-path (the extra part of the path between the base url and internal/management/bands) is empty for a local system, but it is `/facilities/` in the Kubernetes environments.

facilities-tests/src/test/java/gov/va/api/lighthouse/facilities/tests/PssgRequestBot.java
```diff
@@ -0,0 +1,81 @@
...
+  @SneakyThrows
+  public static void main(String[] args) {
+
+    final String BASE_URI = "https://vaww.pssg.med.va.gov/";
+    int offset = 0;
+    final int RECORD_COUNT = 25;
+    PssgBandFeatures response;
+    final SystemDefinitions.Service svc = SystemDefinitions.systemDefinition().facilitiesInternal();
+
+    do {
```
jblefkove-va, line 30: I don't have strong feelings on this since it is not production code, but I think this would be easier to follow as a `while` loop with the exit in the middle:

```
while (true) {
  // call PSSG
  if (pssgResponse.isEmpty()) {
    break;
  }
  // upload bands
}
```

jblefkove-va:
I added support for running from command line: `./drive-time-bandito-rides-again.sh LOCAL axolotl`

facilities-tests/src/test/java/gov/va/api/lighthouse/facilities/tests/PssgRequestBot.java
taylorrassmannlibertyits: Ah gotcha! That makes a lot of sense.

jblefkove-va:
A couple more convenience things:

- Keep a running total of the number of bands processed and print it at the end
- Override the value of the PSSG url to `https://localhost:8666/` if the environment is `LOCAL`

facilities-tests/src/test/java/gov/va/api/lighthouse/facilities/tests/PssgRequestBot.java
```diff
@@ -0,0 +1,88 @@
...
+@Slf4j
+public class PssgRequestBot {
+  private static final ObjectMapper MAPPER = JacksonConfig.createMapper();
+
+  @SneakyThrows
+  public static void main(String[] args) {
+
+    final String BASE_URI =
+        Environment.get() == Environment.LOCAL
+            ? "https://localhost:8666/"
```
jblefkove-va, line 27: http

facilities-tests/src/test/java/gov/va/api/lighthouse/facilities/tests/PssgRequestBot.java
```diff
@@ -0,0 +1,88 @@
...
+        log.info("Post completed!");
+      }
+
+      offset += RECORD_COUNT;
+
+      log.info("New offset: " + offset);
+
+    } while (response.features != null && !response.features.isEmpty());
+
+    log.info("Total PSSG Bands Processed: " + offset);
```
jblefkove-va, line 77: This should be an explicit sum. (If there are 51 total bands this will still tell you 75)

facilities-tests/src/test/java/gov/va/api/lighthouse/facilities/tests/PssgRequestBot.java
```diff
@@ -0,0 +1,88 @@
...
+
+    log.info("Completed PSSG Update!");
```
jblefkove-va, line 79: Maybe combine this with the above log statement, and include the environment? e.g. `Updated 1000 bands in QA`

facilities-tests/src/test/java/gov/va/api/lighthouse/facilities/tests/PssgRequestBot.java
taylorrassmannlibertyits: Ah very true. Will update.

jblefkove-va:
I was able to get total update time down to 90 minutes on my GFE with some optimizations:

- Parallelize requests inside `PssgRequestBot`
- Change `InternalDriveTimeBandController.updateBand()` to accept `PssgResponse` instead of a list of bands, to avoid the cost of deserializing each `PssgResponse` in `PssgRequestBot`
- Parallelize the saving of the bands in `InternalDriveTimeBandController.updateBand()`

PR #140, opened 2020-08-10
API-1823 New endpoint to search facility ids by type

jnguy-va:
https://vajira.max.gov/browse/API-1823

Added a new endpoint, `/facilities/ids?type=health` that searches facilities by type.


jnguy-va:
Not sure if this is the correct endpoint, but I implemented it in this way.
https://sandbox-api.va.gov/services/va_facilities/v0/**facilities/ids?**type=health

Caused a little confusion on the swagger update, not sure how you want to approach that.

facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilitiesController.java
```diff
@@ -351,6 +365,27 @@ GeoFacilitiesResponse geoFacilitiesByState(
         .build();
   }
 
+  /** Get facilities by type. */
```
taylorrassmannlibertyits, line 368: I know this says `Get facilities by type` but the method says `geoFacilitiesByType` maybe add the word "get" at the start of the method name.

facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilitiesController.java
```diff
@@ -521,6 +556,41 @@ FacilitiesResponse jsonFacilitiesByState(
...
```
taylorrassmannlibertyits, line 559: Similar comment to line 368

facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilitiesController.java
```diff
@@ -521,6 +556,41 @@ FacilitiesResponse jsonFacilitiesByState(
...
+      value = "/facilities/ids",
+      produces = "application/json",
+      params = {"!bbox[]", "!lat", "!long", "!state", "!zip", "type"})
+  FacilitiesResponse jsonFacilitiesByType(
+      @RequestParam(value = "type") String type,
+      @RequestParam(value = "services[]", required = false) List<String> services,
+      @RequestParam(value = "page", defaultValue = "1") @Min(1) int page,
+      @RequestParam(value = "per_page", defaultValue = "10") @Min(0) int perPage) {
+    Page<FacilityEntity> entitiesPage =
+        entitiesPageByType(type, services, page, Math.max(perPage, 1));
```
taylorrassmannlibertyits, line 570: Instead of having a Math.max here, could you just set the `@Min(0)` to `@Min(1)` ?

facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilitiesController.java
```diff
@@ -521,6 +556,41 @@ FacilitiesResponse jsonFacilitiesByState(
...
+            .totalEntries((int) entitiesPage.getTotalElements())
+            .build();
+    return FacilitiesResponse.builder()
+        .data(
+            perPage == 0
+                ? emptyList()
+                : entitiesPage.stream().map(e -> facility(e)).collect(toList()))
+        .links(linker.links())
+        .meta(
+            FacilitiesResponse.FacilitiesMetadata.builder().pagination(linker.pagination()).build())
```
taylorrassmannlibertyits, line 590: I think the spacing looks a little off here?

facilities/src/test/java/gov/va/api/lighthouse/facilities/FacilitiesByTypeTest.java
```diff
@@ -0,0 +1,79 @@
...
+  private FacilitiesController controller() {
+    return FacilitiesController.builder()
+        .facilityRepository(repo)
+        .baseUrl("http://foo/")
+        .basePath("bp")
+        .build();
+  }
+
+  @Test
+  void geoFacilitiesbyType() {
```
taylorrassmannlibertyits, line 32: I suggest adding Test to the end of each `@Test` method name.

facilities/src/test/java/gov/va/api/lighthouse/facilities/FacilitiesByTypeTest.java
```diff
@@ -0,0 +1,79 @@
...
+
+  @Test
+  void json_invalidType() {
+    assertThrows(
+        ExceptionsV0.InvalidParameter.class,
+        () -> controller().jsonFacilitiesByType("xxx", null, 1, 1));
+  }
+
+  @Test
+  void jsonperPageZero() {
```
taylorrassmannlibertyits, line 57: Missed the camelCase for "per"

facilities/src/test/java/gov/va/api/lighthouse/facilities/FacilitiesByTypeTest.java
```diff
@@ -0,0 +1,79 @@
+package gov.va.api.lighthouse.facilities;
```
taylorrassmannlibertyits, line 1: I think this test should also test a multi page response. 

I think this test should also test a single page response with multiple entries.

What do you guys think @jnguy-va and @jblefkove-va ?

jblefkove-va:
Unfortunately there was a miscommunication here; the new endpoint needs to return facility IDs, not facilities themselves. See my last entry from the original email thread below:

---

Michael asked earlier in the thread about adding an endpoint to return all facility IDs. Would that be sufficient for this use case?

For example:

Get all health IDs: `https://sandbox-api.va.gov/services/va_facilities/v0/facilities/ids?type=health`

Response: `{"data":["vha_554",...,"vha_636GF"]}`

Filter to IDs containing the desired prefix and use the existing IDs lookup:
`https://sandbox-api.va.gov/services/va_facilities/v0/facilities?ids=vha_528,vha_528A4,...,vha_528QN`

jblefkove-va:
I've updated the Swagger for what the new endpoint should look like. Note the new class `FacilitiesIdsResponse`.

facilities/src/test/java/gov/va/api/lighthouse/facilities/FacilitiesByTypeTest.java
jblefkove-va: We want `Test` as a suffix in our unit test class names so that our tools can detect which files are test suites, but in general it's redundant to include `test` in each individual method name.

jblefkove-va:
Also note there is an existing `findAllIds` method on `FacilityRepository` that can be leveraged for this. (IDs are lightweight so it is sufficient to filter the results of that query, vs. making another query that includes the type filtering.)

facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilitiesIdsController.java
```diff
@@ -0,0 +1,54 @@
...
+
+    for (FacilityEntity ids : allEntities) {
+        filteredByType.add(ids.id().toIdString());
+      }
+    return filteredByType;
+  }
+
+  /** Get all facility IDs as a list by Type. */
+  @GetMapping(
+      value = "/ids",
```
jblefkove-va, line 48: I think it's sufficient to implement this method in `FacilitiesController`. (And note that the parent path on that controller is `/v0`, so the path on this method would be `/facilities/ids`)

facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilitiesIdsController.java
```diff
@@ -0,0 +1,54 @@
...
+      produces = {"application/json"},
+      params = {"type"})
```
jblefkove-va, line 50: I think the `params` tag is unnecessary here. There are a lot of facilities searches that support different combinations of parameters, so we do use these tags to differentiate those. But this is on a different path, so it should not be necessary.

facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilitiesIdsController.java
```diff
@@ -0,0 +1,54 @@
...
+  FacilitiesIdsResponse getFacilityIds(@RequestParam(value = "type", required = false) String type) {
+    return FacilitiesIdsResponse.builder().data(facilityByType(type)).build();
```
jblefkove-va, line 52: `facilityByType(type)` returns a list of strings (facility IDs), so the name should be something like `facilityIdsByType()`

facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilitiesIdsController.java
```diff
@@ -0,0 +1,54 @@
...
+  FacilitiesIdsController(
+      @Autowired FacilityRepository facilityRepository,
+      @Value("${facilities.url}") String baseUrl,
+      @Value("${facilities.base-path}") String basePath) {
+    this.facilityRepository = facilityRepository;
+  }
+
+  List<String> facilityByType(String rawType) {
+    List<String> filteredByType = new ArrayList<>();
+    FacilityEntity.Type facilityType = validateFacilityType(rawType);
```
jblefkove-va, line 32: With the valid facility type, you should be able to directly filter the results of `facilityRepository.findAllIds()`. Something like:

```
List<FacilityEntity.Pk> ids =
      facilityRepository
          .findAllIds()
          .stream()
          .filter(id -> facilityType == null || id.type() == facilityType)
          .collect(toList());
```

facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilitiesController.java
```diff
@@ -268,6 +270,19 @@ private FacilityEntity entityById(String id) {
...
 
+  List<String> facilityIdsByType(String rawType) {
+    List<String> filteredByType = new ArrayList<>();
+    FacilityEntity.Type facilityType = validateFacilityType(rawType);
+    List<FacilityEntity.Pk> ids =
+        facilityRepository.findAllIds().stream()
+            .filter(id -> facilityType == null || id.type() == facilityType)
+            .collect(toList());
+    for (FacilityEntity.Pk id : ids) {
+      filteredByType.add(id.toIdString());
```
jblefkove-va, line 281: you can lump this operation into the above stream as well. (After filtering, before `collect(toList())`: `.map(id -> id.toIdString())`

facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilitiesController.java
```diff
@@ -558,6 +573,15 @@ FacilitiesResponse jsonFacilitiesByZip(
         .build();
   }
 
+  /** Get all facility IDs as a list by Type. */
+  @GetMapping(
+      value = "/facilities/ids",
+      produces = {"application/json"})
+  FacilitiesIdsResponse jsonFacilityIdsByType(
```
jblefkove-va, line 580: The `json` prefix on the methods in this file is for differentiating methods that return results in regular format vs. those that use GeoJSON. Since this is just the single method and we only support one format, the `json` prefix is unnecessary.

facilities/src/test/java/gov/va/api/lighthouse/facilities/FacilitiesIdsControllerTest.java
```diff
@@ -0,0 +1,46 @@
...
+
+  @Test
+  void facilityIdsByType() {
+    repo.save(FacilitySamples.defaultSamples().facilityEntity("vha_757"));
+    assertThat(controller().facilityIdsByType("health"))
+        .isEqualTo(List.of(FacilitySamples.defaultSamples().facility("vha_757").id()));
+  }
+
+  @Test
+  void jsonFacilityIdsByType() {
```
jblefkove-va, line 36: Some other test cases to add:

- A valid type that doesn't match the test facility, so no results are returned
- Null type, so all results are returned

facilities/src/test/resources/search-ids-type.json
```diff
@@ -0,0 +1,5 @@
+{
+  "data" : [
```
jblefkove-va, line 2: This file appears to be unused?

facilities-tests/src/test/java/gov/va/api/lighthouse/facilities/tests/SearchIT.java
```diff
@@ -83,6 +84,13 @@ void searchByIds() {
     facilitiesRequest(null, request, 200).expectValid(FacilitiesResponse.class);
   }
 
+  @Test
+  void searchByIdsByType() {
+    final String request = "v0/facilities/ids?type=";
```
jblefkove-va, line 89: The endpoints in `FacilitiesController` are split across three IT files: `SearchIT` is the location-based searches, while `ReadIT` has the reads and `BulkIT` is the `/all` endpoint. So I think the `/ids` endpoint should have its own file as well. (And there should be separate tests for `/ids` with no type vs `/ids?type=health`)

PR #142, opened 2020-08-12
F: NullPointerException in testEtlLastUpdated 

taylorrassmannlibertyits:
https://vajira.max.gov/browse/API-1999
Updated for null checking, and unit test coverage.


facilities/src/main/java/gov/va/api/lighthouse/facilities/HealthController.java
```diff
@@ -211,11 +211,19 @@ private Health testEtlLastUpdated(@NonNull Instant now) {
     Instant twentyFourHoursEarlier = now.minus(24, ChronoUnit.HOURS);
     String statusCode =
         lastUpdated != null && !lastUpdated.isBefore(twentyFourHoursEarlier) ? "UP" : "DOWN";
-    return Health.status(new Status(statusCode, "ETL process succeeded in the last 24 hours"))
-        .withDetail("name", "Facilities ETL process")
-        .withDetail("time", now)
-        .withDetail("lastUpdated", lastUpdated)
-        .build();
+
+    if (lastUpdated == null) {
```
jblefkove-va, line 215: Let's move this up and bail out immediately if `lastUpdated` is null. (And then the other null check for `lastUpdated` on line 212 becomes redundant and can be removed.)

taylorrassmannlibertyits: Since `statusCode` is used in constructing the return value we will either need to copy/paste `statusCode` twice and remove the null check, or we could leave as is?

jblefkove-va: You can remove the null check and hardcode status code as `DOWN` for the case where `lastUpdated` is `null`

taylorrassmannlibertyits: Ah that is good! Will update!

PR #144, opened 2020-08-18
Api 1757 get benefits centers from cdw

taylorrassmannlibertyits:
https://vajira.max.gov/browse/API-1757
Updated to remove arcgis benefits retrieval to added CDW benefits centers retrieval.


facilities-local-db/src/test/java/gov/va/api/lighthouse/facilitieslocaldb/Populaterator.java
```diff
@@ -257,6 +258,40 @@ private void vast(Connection connection) {
...
+      connection
+          .prepareStatement(
+              "CREATE TABLE App.FacilityLocator_VBA ("
+                  + headers.stream().map(h -> h + " VARCHAR(MAX)").collect(joining(","))
+                  + ")")
+          .execute();
+      for (CSVRecord row : parser) {
+        try (PreparedStatement statement =
+            connection.prepareStatement(
+                "INSERT INTO App.FacilityLocator_VBA ("
```
jblefkove-va, line 278: This could be a little more readable with `String.format()`: `INSERT INTO App.FacilityLocator_VBA (%s) VALUES (%s)`

facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/BenefitsCollector.java
```diff
@@ -3,38 +3,120 @@
...
 @Slf4j
 final class BenefitsCollector {
-  private final String arcgisUrl;
-
   private final Map<String, String> websites;
 
-  private final RestTemplate restTemplate;
+  private final JdbcTemplate jdbcTemplate;
+
+  /** Convert the results into attributes and geometry to make a CdwBenefits Object. */
```
jblefkove-va, line 25: the comment is out of date

facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/BenefitsCollector.java
```diff
@@ -3,38 +3,120 @@
...
+        .phone(rs.getString("PHONE"))
+        .monday(rs.getString("MONDAY"))
+        .tuesday(rs.getString("TUESDAY"))
+        .wednesday(rs.getString("WEDNESDAY"))
+        .thursday(rs.getString("THURSDAY"))
+        .friday(rs.getString("FRIDAY"))
+        .saturday(rs.getString("SATURDAY"))
+        .sunday(rs.getString("SUNDAY"))
+        .applyingForBenefits(
+            rs.getString("APPLYING_FOR_BENEFITS") == null
```
jblefkove-va, line 47: I think it's safe to pass the value in directly here, even if it's null. The `@Builder.default` on the `CdwBenefits` object will kick-in in that case.

facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/CdwBenefits.java
```diff
@@ -0,0 +1,157 @@
+package gov.va.api.lighthouse.facilities.collector;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
```
jblefkove-va, line 4: The jackson annotations can be removed from this file.

facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/CdwBenefits.java
```diff
@@ -0,0 +1,157 @@
...
+  String websiteUrl;
+
+  @JsonProperty("x")
+  BigDecimal longitude;
+
+  @JsonProperty("y")
+  BigDecimal latitude;
+
+  /* Unused Fields:
+   * objectIdFieldName
```
jblefkove-va, line 137: This comment about the unused fields can be removed. (These were present in the results from ArcGIS, but they are not mapped in our CDW dataset.)

facilities/src/test/java/gov/va/api/lighthouse/facilities/collector/BenefitsCollectorTest.java
```diff
@@ -3,48 +3,62 @@
 import static java.util.Collections.emptyMap;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.jupiter.api.Assertions.assertThrows;
-import static org.mockito.ArgumentMatchers.eq;
-import static org.mockito.ArgumentMatchers.startsWith;
+import static org.mockito.ArgumentMatchers.*;
```
jblefkove-va, line 6: Starred imports are frowned upon. (This one isn't flagged by static analysis since it's a test.) Check your IDE settings to disable this.

facilities/src/test/java/gov/va/api/lighthouse/facilities/collector/BenefitsCollectorTest.java
```diff
@@ -3,48 +3,62 @@
...
   @SneakyThrows
   void collect() {
-    RestTemplate restTemplate = mock(RestTemplate.class);
-
-    ResponseEntity<String> body =
-        ResponseEntity.of(
-            Optional.of(
-                new String(
-                    getClass().getResourceAsStream("/arcgis-benefits.json").readAllBytes())));
+    JdbcTemplate jdbcTemplate = mock(JdbcTemplate.class);
```
jblefkove-va, line 26: Take a look at `InternalCollectorControllerTest`. It uses an autowired JdbcTemplate. (There is still some set-up involved in manually creating your table and inserting a row into it, but overall it is probably still easier to follow than mocked jdbc templates and result sets.)

facilities/src/test/java/gov/va/api/lighthouse/facilities/collector/BenefitsCollectorTest.java
```diff
@@ -103,15 +117,34 @@ void collect() {
...
-                .arcgisUrl("http://wrong:8080")
-                .restTemplate(restTemplate)
-                .websites(emptyMap())
-                .build()
-                .collect());
+        () -> BenefitsCollector.builder().websites(emptyMap()).build().collect());
+  }
+
+  @Test
+  void toCdwBenefits() throws SQLException {
```
jblefkove-va, line 126: `@SneakyThrows`

facilities/src/test/java/gov/va/api/lighthouse/facilities/collector/BenefitsTransformerTest.java
```diff
@@ -7,63 +7,36 @@
...
-                        .build()))
-        .isEqualTo(facilityService(Facility.BenefitsService.UpdatingDirectDepositInformation));
-    assertThat(
-            tx().services(
-                    ArcGisBenefits.Attributes.builder()
-                        .vocationalRehabilitationEmplo("YES")
-                        .build()))
+
+    // Note the BenefitsSamples class defines the valid facility services for this test. AND WE GET
+    // THEM ALL!!!
```
jblefkove-va, line 15: image

facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/BenefitsCollector.java
```diff
@@ -43,39 +125,56 @@
     }
   }
 
-  /** Requests ArcGIS VA_Benefits_Facilities in application/json. */
+  /** Requests CDW VA_Benefits_Facilities in application/json. */
```
jblefkove-va, line 128: I don't think the `application/json` comment is still applicable since this is no longer a web service call.

facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/BenefitsCollector.java
taylorrassmannlibertyits: Updated in next commit.

facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/BenefitsCollector.java
taylorrassmannlibertyits: Unfortunately the code uses a deeper method for determining benefits:
```
private boolean yesnoToBoolean(String yesNo) {
    /* Assume no if the answer is not an emphatic yes. */
    return yesNo.equalsIgnoreCase("YES");
  }
```
I can modify this method to handle null values (which should just be NO)

jblefkove-va: My mistake, the builder defaults are not used if you explicitly override them as null. In that case updating the transformer to gracefully handle null is the best option.

facilities/src/test/java/gov/va/api/lighthouse/facilities/collector/BenefitsCollectorTest.java
taylorrassmannlibertyits: Ah will do! I always rely on the static analysis to check me.

facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/BenefitsCollector.java
taylorrassmannlibertyits: Agreed

facilities/src/test/java/gov/va/api/lighthouse/facilities/collector/BenefitsCollectorTest.java
taylorrassmannlibertyits: Yea I had some "fun" times making the mocked versions. I'll look into the autowired stuffs and see how it goes.

taylorrassmannlibertyits: At this time, I think I'm going to leave as is. I took a look at the `InternalCollectorControllerTest` and see its usefulness for the future.

facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/BenefitsTransformer.java
```diff
@@ -12,115 +12,117 @@
...
   }
 
   private boolean yesnoToBoolean(String yesNo) {
-    /* Assume no if the answer is not an emphatic yes. */
-    return yesNo.equalsIgnoreCase("YES");
+    /* Assume no if the answer is not an emphatic yes or the string is null. */
+    if (yesNo == null) {
+      return false;
+    } else {
+      return yesNo.equalsIgnoreCase("YES");
```
jblefkove-va, line 125: Instead of the conditional, how about `"YES".equalsIgnoreCase(yesNo)`

facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/BenefitsTransformer.java
```diff
@@ -12,115 +12,117 @@
...
-    /* ArcGIS returns a string NULL... We don't want to return that.*/
-    return arcgisWebsite == null || arcgisWebsite.equalsIgnoreCase("NULL")
-        ? csvWebsite
-        : arcgisWebsite;
+  String website(String website) {
+    /* CDW returns a string NULL... We don't want to return that.*/
+    return website == null || website.equalsIgnoreCase("NULL") ? csvWebsite : website;
   }
 
   private boolean yesnoToBoolean(String yesNo) {
```
jblefkove-va, line 120: While we're in here, fix the camelCasing on `yesno`

facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/CdwBenefits.java
```diff
@@ -0,0 +1,83 @@
...
+
+  String thursday;
+
+  String friday;
+
+  String saturday;
+
+  String sunday;
+
+  @Builder.Default String applyingForBenefits = "NO";
```
jblefkove-va, line 44: Since we've updated the transformer to handle nulls, are these builder defaults still necessary?

facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/CdwBenefits.java
```diff
@@ -0,0 +1,83 @@
...
+
+  String otherServices;
+
+  String websiteUrl;
+
+  BigDecimal longitude;
+
+  BigDecimal latitude;
+
+  static final class CdwBenefitsBuilder {}
```
jblefkove-va, line 82: You can remove this. The reason it was present in the old `ArcGisBenefits` class was so that we could put a `JsonIgnore` annotation on it.

For background, we have Jackson configuration in health-apis-parent that uses Lombok builders for deserialization, when available, for immutable classes. Out of the box, Jackson cannot deserialize into POJOs that have final fields or no setters. See here: https://github.com/department-of-veterans-affairs/health-apis-parent/blob/master/service-auto-config/src/main/java/gov/va/api/health/autoconfig/configuration/JacksonConfig.java#L120

facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/CdwBenefits.java
taylorrassmannlibertyits: I'll remove and test it out, but I think you're right, they probably aren't necessary anymore.

facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/BenefitsTransformer.java
taylorrassmannlibertyits: Me gusta

facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/CdwBenefits.java
taylorrassmannlibertyits: Unfortunately it leaves the values as null here, so I'm going to keep the defaults

jblefkove-va: Right, but I think it's okay for these to be null now. (You may have to adjust some unit tests, but the transformer is good.)

taylorrassmannlibertyits: Oh I see what you mean! Sounds good

PR #147, opened 2020-08-21
Sanitize <Null> addresses

taylorrassmannlibertyits:
https://vajira.max.gov/browse/API-2104
Added code to filter out <Null> from addresses.


facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/Transformers.java
```diff
@@ -25,6 +25,14 @@ static boolean allBlank(Object... values) {
     return true;
   }
 
+  static String checkAddressForNullKeyword(String address) {
```
jblefkove-va, line 28: Even though we only use it for street addresses, this method itself isn't specific to that. How about `checkAngleBracketNull()`?

facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/HealthTransformer.java
```diff
@@ -256,9 +257,9 @@ private String id() {
         .zip(zip())
         .city(vast.city())
         .state(upperCase(vast.state(), Locale.US))
-        .address1(vast.address2())
-        .address2(vast.address1())
-        .address3(vast.address3())
+        .address1(checkAddressForNullKeyword(vast.address2()))
```
jblefkove-va, line 260: these checks should also be done in the call to `allBlank()` above

facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/StateCemeteryTransformer.java
```diff
@@ -37,8 +38,8 @@
           .zip(parseZip(line3))
           .city(parseCity(line3))
           .state(state)
-          .address1(line1)
-          .address2(line2)
+          .address1(checkAddressForNullKeyword(line1))
```
jblefkove-va, line 41: These variables are referenced in several places through this method. I think it would be cleaner to move these checks to the two different places where the `asAddress()` method is invoked, so that the data is already sanitized when it enters this method. (We do something similar forcing the `state` parameter to uppercase.)

facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/VetCenterTransformer.java
```diff
@@ -41,8 +42,8 @@
         .zip(addressZip())
         .city(vast.city())
         .state(upperCase(vast.state(), Locale.US))
-        .address1(vast.address2())
-        .address2(vast.address3())
+        .address1(checkAddressForNullKeyword(vast.address2()))
```
jblefkove-va, line 45: Ditto for also checking `allBlank()` above

facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/Transformers.java
taylorrassmannlibertyits: Naming is never my strong suit. I like this better :)

facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/HealthTransformer.java
taylorrassmannlibertyits: Ah I misinterpreted all blank as empty, but makes sense since `"<Null>"` is no bueno. Checks incoming!

facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/Transformers.java
```diff
@@ -25,6 +25,14 @@ static boolean allBlank(Object... values) {
     return true;
   }
 
+  static String checkAngleBracketNull(String address) {
```
jblefkove-va, line 28: Parameter should also be renamed 😄

jblefkove-va, line 28: Add a transformers unit test for this method

jnguy-va (dismissed):
Local tests look good currently, if you want me to rerun again let me know.

PR #150, opened 2020-08-27
API-2209 protect against empty responses during collection

jblefkove-va:
https://vajira.max.gov/browse/API-2209


facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/CemeteriesCollector.java
```diff
@@ -79,6 +80,7 @@ private ArcGisCemeteries requestArcGisCemeteries() {
         "Loading cemeteries took {} millis for {} features",
         totalWatch.stop().elapsed(TimeUnit.MILLISECONDS),
         cemeteries.features().size());
+    checkState(!cemeteries.features().isEmpty(), "No ArcGIS cemeteries");
```
taylorrassmannlibertyits, line 83: Depending on when @jnguy-va 's PR goes in, do you want this to say CDW?

facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/StateCemeteriesCollector.java
```diff
@@ -50,27 +50,32 @@
...
+        new XmlMapper()
+            .registerModule(new StringTrimModule())
+            .disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES)
+            .readValue(response, StateCemeteries.class)
+            .cem();
+    log.info(
+        "Loading non-national cemeteries took {} millis for {} entries",
+        totalWatch.stop().elapsed(TimeUnit.MILLISECONDS),
+        cemeteries.size());
+    checkState(!cemeteries.isEmpty(), "No cems.xml entries");
```
taylorrassmannlibertyits, line 77: I think this log is correct, I just feel something funny about this section. Maybe `checkState` should occur before the log? I see the log will denote there are 0 entries though, so I guess this comment is to open a discussion.

facilities/src/test/java/gov/va/api/lighthouse/facilities/collector/FacilitiesCollectorTest.java
```diff
@@ -9,41 +9,157 @@
...
 public class FacilitiesCollectorTest {
-  @Test
-  void toVastEntity() {
-    assertThat(FacilitiesCollector.toVastEntity(mock(ResultSet.class)))
-        .isEqualTo(
-            VastEntity.builder().vetCenter(false).mobileVetCenter(false).mobile(false).build());
+  @Autowired JdbcTemplate jdbcTemplate;
+
+  private void _initDatabase() {
+    jdbcTemplate.execute(
```
taylorrassmannlibertyits, line 35: Thank you for making this. 😃

facilities/src/test/java/gov/va/api/lighthouse/facilities/collector/FacilitiesCollectorTest.java
```diff
@@ -78,20 +210,18 @@ void verifyResponse() {
...
-                    mock(JdbcTemplate.class),
+                    jdbcTemplate,
                     restTemplate,
                     "http://arcgis",
                     "http://atc",
                     "http://atp",
                     "http://statecems")
                 .collectFacilities())
-        .isEmpty();
+        .isNotEmpty();
```
taylorrassmannlibertyits, line 225: Since we are no longer empty, should there be any other tests here looking at what is collected? Or do the other test classes verify this enough for us to be happy?

facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/CemeteriesCollector.java
jblefkove-va: This whole method is going to disappear as part of the switchover to CDW, so we will just want to make sure that the new one in her PR includes a similar check.

facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/StateCemeteriesCollector.java
jblefkove-va: I put the check after the log across the board so that there will be an explicit log message indicating "0 entries" in addition to the exception stack trace. (I moved stuff around in here because this collector was doing the timing differently from the others. The other collectors put the timing only around the web service request or the database call, not the full transformation to our `Facility` model. This collector was previously putting the log and timing at the very end.)

facilities/src/test/java/gov/va/api/lighthouse/facilities/collector/FacilitiesCollectorTest.java
jblefkove-va: The way we interact with the jdbcTemplate to populate mental health numbers and stop codes cannot easily be mocked out, since it uses a `RowCallbackHandler` with no return type.

facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/StateCemeteriesCollector.java
taylorrassmannlibertyits: ok cool thanks! Funny feeling removed!

facilities/src/test/java/gov/va/api/lighthouse/facilities/collector/FacilitiesCollectorTest.java
jblefkove-va: I changed it to check explicit size. But beyond that I think the unit tests for the individual collector classes cover it.

PR #151, opened 2020-08-28
Swagger Update for deprecations and new notes on formats and constraints

taylorrassmannlibertyits:
https://vajira.max.gov/browse/API-2083
Updated swagger to consolidate deprecations and add notes for facility ID formats and constraints.


facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/FacilitiesService.java
```diff
@@ -74,14 +72,28 @@
...
+                    + "A station ID has the format prefix_stationNumber. The prefix is one "
+                    + "of nca, vc, vba, or vha. Cemeteries may be national (VA) or "
+                    + "non-national; non-national cemeteries have the station number prefixed "
+                    + "with an 's' character. There are no other constraints on the format.\n"
+                    + "Examples:\n"
+                    + "- National cemetery: nca_063\n"
+                    + "- Non-national cemetery: nca_s1082\n"
+                    + "- Vet center: vc_0872MVC\n"
+                    + "- Benefits: vba_539GB\n"
+                    + "- Health: vha_402GA\n"
```
jblefkove-va, line 87: Let's add the prefixes up on lines 28-31 so it is clear from the outset what facility types they represent.

jblefkove-va:
There is a note in `Facility.java` about `active_status` being removed in version 1.0.0.

facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/FacilitiesService.java
```diff
@@ -58,8 +58,6 @@
                     + "- `application/geo+json` "
                     + "- GeoJSON-compliant format, representing each facility as a "
                     + "feature with a point geometry.\n"
-                    + "- `application/vnd.geo+json` "
-                    + "- Deprecated. Will be replaced by `application/geo+json` in version 1.0.0.\n"
```
jblefkove-va, line 62: Since we do still support `vnd.geo+json`, I think this bullet should remain. But the description can be shorted to something like "Deprecated. Prefer `application/geo+json`."

taylorrassmannlibertyits: Is it ok for me to move to the new deprecated section, or would you think it's better here?

taylorrassmannlibertyits: Ah since it's a response format I think staying here but saying deprecated is the way to go.

facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/FacilitiesService.java
```diff
@@ -74,14 +72,28 @@
...
+                    + "- Health: vha_402GA\n"
                     + "\n\n### Mobile Facilities\n\n"
                     + "The mobile health facilities move regularly within a region. "
                     + "If a facility comes back from this API with `\"mobile\": \"true\"`, "
                     + "the latitude/longitude and address could be inaccurate. "
                     + "To get the exact current location, please call the number listed."
                     + "\n\n## Deprecations\n\n"
-                    + "- `active_status` is deprecated and replaced with `operating_status`."
-                    + " It will be removed in version 1.0.0."
+                    + "- `active_status` is deprecated and replaced with `operating_status`.\n"
```
jblefkove-va, line 94: Let's clarify `active_status` *field* is deprecated and `application/vnd.geo+json` *media type* is deprecated

facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/FacilitiesService.java
```diff
@@ -74,14 +72,28 @@
                     + "are only applicable to\n  VA health facilities.\n"
                     + "- The list of available services in the `services` element is only "
                     + "applicable to VA health and\n  benefits facilities."
+                    + "\n\n"
+                    + "### Facility ID Formats and Constraints"
+                    + "\n\n"
+                    + "A station ID has the format prefix_stationNumber. The prefix is one "
```
jblefkove-va, line 78: Should be facility ID, not station ID. Also let's add backticks around `prefix_stationNumber` here, and the example IDs below.

facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/FacilitiesService.java
jblefkove-va, line 87: Change the order here (health, benefits, cems, vet centers) to match the order of types at the beginning.

facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/FacilitiesService.java
```diff
@@ -74,14 +72,28 @@
...
+                    + "of nca, vc, vba, or vha. Cemeteries may be national (VA) or "
+                    + "non-national; non-national cemeteries have the station number prefixed "
+                    + "with an 's' character. There are no other constraints on the format.\n"
```
jblefkove-va, line 81: This can be "prefixed with an `s`" with backticks and drop `character`.

jblefkove-va, line 81: The newline at the end may be redundant? When I view in Swagger editor I see `There are no other constraints on the format. Examples:` all as one line, which seems okay.

PR #155, opened 2020-09-11
Missing Fields Notifications in Facilities

taylorrassmannlibertyits:
https://vajira.max.gov/browse/API-2317
Updated to warn against missing values:
1. Empty Phone numbers from facilities (only notify if main number is blank for now)
2. Empty Street Addresses
3. Empty Zip
4. Empty City
5. empty State
6. Empty Lat/Lon
7. Empty Hours (any day of week)
8. Empty Classification
9. Empty Website
10. services.health is empty for a health facility
11. services.benefits is empty for a benefits center


facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
```diff
@@ -55,6 +58,31 @@
 
   private final FacilityGraveyardRepository graveyardRepository;
 
+  static boolean allBlank(Object... values) {
```
taylorrassmannlibertyits, line 61: @jblefkove-va I took these two methods from `Transformer.java` but I'm thinking maybe we should move that UtilitiesClass up one level? Or maybe move these two methods into their own class?

jblefkove-va: Classes in the top-level package can reference stuff in the `facilities.collector` package, so you can just change `Transformers` visibility to public

jblefkove-va:
After our conversation on friday, I realized a way to test this with real data. Hit the internal collection endpoint `https://blue.qa.lighthouse.va.gov/facilities/internal/collector/facilities` from GFE, which does the collection process but does not reload anything. Then locally use the debug version of the reload endpoint to which you can POST a collector response: 

```
curl -s -X POST -H "Content-Type: application/json" -H "client-key: axolotl" -d @collect.json http://localhost:8085/internal/management/reload
```

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
```diff
@@ -306,6 +334,91 @@ void updateAndSave(ReloadResponse response, FacilityEntity record, Facility faci
...
+        response.problems().add(ReloadResponse.Problem.of(facility.id(), "Missing city"));
+      }
+    } else {
+      response
+          .problems()
+          .add(
+              ReloadResponse.Problem.of(
+                  facility.id(), "Missing all physical " + "address information"));
+    }
+    if (allBlank(facility.attributes().address().mailing())) {
```
jblefkove-va, line 348: The mailing address object could be present with all of its fields null

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
```diff
@@ -306,6 +334,91 @@ void updateAndSave(ReloadResponse response, FacilityEntity record, Facility faci
     if (isBlank(record.state())) {
       response.problems().add(ReloadResponse.Problem.of(facility.id(), "Missing state"));
     }
+    if (facility.attributes().address().physical() != null) {
+      if (isBlank(facility.attributes().address().physical().city())) {
```
jblefkove-va, line 338: Take a look at `CsvTransformer`. There is a pattern in there for using Optionals for these deep traversals. You should be able to do this in one check with something like `isBlank(addressPhysical(facility).map(a -> a.city()))`

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
```diff
@@ -306,6 +334,91 @@ void updateAndSave(ReloadResponse response, FacilityEntity record, Facility faci
...
+          .add(ReloadResponse.Problem.of(facility.id(), "Missing all hours information"));
+    }
+
+    if (isBlank(facility.attributes().classification())) {
+      response
+          .problems()
+          .add(ReloadResponse.Problem.of(facility.id(), "Missing facility classification"));
+    }
+    if (isBlank(facility.attributes().website())) {
+      response.problems().add(ReloadResponse.Problem.of(facility.id(), "Missing facility website"));
```
jblefkove-va, line 396: `Missing website` or `Missing classification` is probably sufficient

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
```diff
@@ -306,6 +334,91 @@ void updateAndSave(ReloadResponse response, FacilityEntity record, Facility faci
...
+      response.problems().add(ReloadResponse.Problem.of(facility.id(), "Invalid longitude"));
+    }
+
+    if (facility.attributes().facilityType() == Facility.FacilityType.va_health_facility
+        && allBlank(facility.attributes().services().health())) {
+      response
+          .problems()
+          .add(
+              ReloadResponse.Problem.of(
+                  facility.id(), "Missing health services for health facility"));
```
jblefkove-va, line 411: I think `Missing services` is sufficient, the type can be inferred

facilities/src/test/java/gov/va/api/lighthouse/facilities/InternalFacilitiesControllerTest.java
```diff
@@ -45,6 +45,39 @@
 
   FacilitiesCollector collector = mock(FacilitiesCollector.class);
 
+  private static Facility _facility(
```
jblefkove-va, line 48: Now that the NullPointerExceptions are fixed, is this still necessary? It's probably sufficient to just update the two tests that examine the list of problems (`collect_noStateOrZip` and `updateAndSave_error`) to include the new problems that are emitted for their incomplete test data.

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
taylorrassmannlibertyits: Updated and combined conditionals into one for this `ReloadResponse`

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
taylorrassmannlibertyits: At least that is my intention in this conditional

jblefkove-va: `allBlank` is only checking if `address().mailing()` is null. It's not going to do any introspection on the fields of `mailing()`. `address().mailing()` could be present with all of its fields null, which is still no address information, but that won't trigger a warning.

facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/Transformers.java
```diff
@@ -15,8 +15,10 @@
 import org.apache.commons.lang3.StringUtils;
 
 @UtilityClass
-final class Transformers {
-  static boolean allBlank(Object... values) {
+public final class Transformers {
+
+  /** Check if each value of an object is blank. */
```
jblefkove-va, line 20: That javadoc is misleading, it should be "check if all values are blank"

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
```diff
@@ -296,17 +311,102 @@ void updateAndSave(ReloadResponse response, FacilityEntity record, Facility faci
...
+      response.problems().add(ReloadResponse.Problem.of(facility.id(), "Missing website"));
+    }
+    if (record.latitude() > 90 || record.latitude() < -90) {
+      response.problems().add(ReloadResponse.Problem.of(facility.id(), "Invalid latitude"));
+    }
+    if (record.longitude() > 180 || record.longitude() < -180) {
+      response.problems().add(ReloadResponse.Problem.of(facility.id(), "Invalid longitude"));
+    }
+    if ((facility.attributes().facilityType() == Facility.FacilityType.va_health_facility
+            && allBlank(facility.attributes().services().health()))
```
jblefkove-va, line 405: `facility.attributes().services()` could be null

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
```diff
@@ -112,6 +113,22 @@ static String zipOf(Facility facility) {
     return null;
   }
 
+  private Optional<Facility.Address> addressMailing(Facility facility) {
```
jblefkove-va, line 116: These helper methods can be static

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
```diff
@@ -296,17 +311,102 @@ void updateAndSave(ReloadResponse response, FacilityEntity record, Facility faci
...
+      response
+          .problems()
+          .add(ReloadResponse.Problem.of(facility.id(), "Missing physical address state"));
+    }
+    if (isBlank(addressPhysical(facility).map(a -> a.city()))) {
+      response
+          .problems()
+          .add(ReloadResponse.Problem.of(facility.id(), "Missing physical address city"));
+    }
+    if (isBlank(addressMailing(facility).map(a -> a.zip()))) {
```
jblefkove-va, line 332: For physical and mailing addresses, there should also be a check for the street lines. At least one of `address1`, `address2`, `address3` should not be blank.

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
```diff
@@ -296,17 +311,102 @@ void updateAndSave(ReloadResponse response, FacilityEntity record, Facility faci
     populate(record, facility);
     record.missingTimestamp(null);
     record.lastUpdated(response.timing().completeCollection());
-
     /*
      * Determine if there is something wrong with the record, but it is still usable.
      */
     if (isBlank(record.zip())) {
```
jblefkove-va, line 317: There is a subtle difference here in how zip code is handled for physical address vs. mailing address now. The `zipOf` method that is used to populate the entity zip field verifies that the physical zip has at least five characters, and populates the zip using the first five characters. Thus, a malformed zip of four characters will trigger a missing warning for physical address, but not for mailing address. 

Validating zip format (5 or 5-4 format) is beyond the scope of this story. To make the behavior consistent, `zipOf` should be updated to remove the minimum length requirement and populate up to the first five characters: `zip.substring(0, Math.min(5, zip.length()))` instead of `substring(0,5)`

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
```diff
@@ -296,17 +311,102 @@ void updateAndSave(ReloadResponse response, FacilityEntity record, Facility faci
...
+      response
+          .problems()
+          .add(ReloadResponse.Problem.of(facility.id(), "Missing mailing address state"));
+    }
+    if (isBlank(addressMailing(facility).map(a -> a.city()))) {
+      response
+          .problems()
+          .add(ReloadResponse.Problem.of(facility.id(), "Missing mailing address city"));
+    }
+    if (allBlank(facility.attributes().address().mailing())) {
```
jblefkove-va, line 347: Now that the address fields are handled individually, this check is no longer necessary.

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
```diff
@@ -296,17 +311,102 @@ void updateAndSave(ReloadResponse response, FacilityEntity record, Facility faci
...
+      response
+          .problems()
+          .add(
+              ReloadResponse.Problem.of(
+                  facility.id(), "Missing all " + "mailing address information"));
+    }
+    if (facility.attributes().phone() != null) {
+      if (isBlank(facility.attributes().phone().main())) {
```
jblefkove-va, line 355: This check should be handled similar to the address checks above, so that there is the same behavior for both `phone()` being null and `phone()` being non-null while `phone().main()` is null.

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
```diff
@@ -296,17 +311,102 @@ void updateAndSave(ReloadResponse response, FacilityEntity record, Facility faci
...
+        response
+            .problems()
+            .add(ReloadResponse.Problem.of(facility.id(), "Missing main phone number"));
+      }
+    } else {
+      response.problems().add(ReloadResponse.Problem.of(facility.id(), "No phone numbers present"));
+    }
+    if (facility.attributes().hours() != null) {
```
jblefkove-va, line 363: ditto the above comment for phone

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
```diff
@@ -296,17 +311,102 @@ void updateAndSave(ReloadResponse response, FacilityEntity record, Facility faci
...
+      }
+    } else {
+      response
+          .problems()
+          .add(ReloadResponse.Problem.of(facility.id(), "Missing all hours information"));
+    }
+    if (isBlank(facility.attributes().classification())) {
+      response.problems().add(ReloadResponse.Problem.of(facility.id(), "Missing classification"));
+    }
+    if (isBlank(facility.attributes().website())) {
```
jblefkove-va, line 395: Running this code against the QA data, there is a lot of noise in the warnings list. Three things jump out at me:

- Mailing address is (surprisingly) only populated for cemeteries. So that check should be restricted only to cemeteries.
- Similarly, classification is not populated for vet centers.
- Finally, there are 1200+ facilities with no website populated. The main source of website URLs is the websites CSV document that we own, in this repo. Given how much data is missing, I think the website check should be deferred to a separate story, to be done after we have refreshed the website data with a more complete set.

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
```diff
@@ -296,17 +315,107 @@ void updateAndSave(ReloadResponse response, FacilityEntity record, Facility faci
...
+          .problems()
+          .add(ReloadResponse.Problem.of(facility.id(), "Missing physical address city"));
+    }
+    if (allBlank(
+        addressPhysical(facility).map(a -> a.address1()),
+        addressPhysical(facility).map(a -> a.address2()),
+        addressPhysical(facility).map(a -> a.address3()))) {
+      response
+          .problems()
+          .add(ReloadResponse.Problem.of(facility.id(), "Missing all physical street addresses"));
```
jblefkove-va, line 342: "Missing physical address street"?

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
```diff
@@ -296,17 +315,107 @@ void updateAndSave(ReloadResponse response, FacilityEntity record, Facility faci
...
+    }
+
+    // Mailing addresses currently only exist for cemeteries
```
jblefkove-va, line 345: Nitpick, but I would drop `currently`, there's not anything on the horizon for populating mailing address for the other facility types.

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
```diff
@@ -296,17 +315,107 @@ void updateAndSave(ReloadResponse response, FacilityEntity record, Facility faci
...
+    if (isHoursNull(facility) || isBlank(facility.attributes().hours().saturday())) {
+      response.problems().add(ReloadResponse.Problem.of(facility.id(), "Missing Saturday hours"));
+    }
+    if (isHoursNull(facility) || isBlank(facility.attributes().hours().sunday())) {
+      response.problems().add(ReloadResponse.Problem.of(facility.id(), "Missing Sunday hours"));
+    }
+
+    // Currently classification is not populated for vet centers
+    if (facility.attributes().facilityType() != Facility.FacilityType.vet_center) {
+      if (isBlank(facility.attributes().classification())) {
```
jblefkove-va, line 400: would it be simpler to combine the conditionals since there's only one inner

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
```diff
@@ -296,17 +315,107 @@ void updateAndSave(ReloadResponse response, FacilityEntity record, Facility faci
...
+      }
     }
 
+    if (record.latitude() > 90 || record.latitude() < -90) {
+      response.problems().add(ReloadResponse.Problem.of(facility.id(), "Invalid latitude"));
+    }
+    if (record.longitude() > 180 || record.longitude() < -180) {
+      response.problems().add(ReloadResponse.Problem.of(facility.id(), "Invalid longitude"));
+    }
+    if (facility.attributes().services() != null) {
```
jblefkove-va, line 411: if `services()` is null, then we are missing services for health/benefits

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
jblefkove-va: To clarify I was suggesting that last change for this PR. (The rest, e.g. indicating that a zip code with 4 characters is not valid, is a separate story.)

taylorrassmannlibertyits: Ah ok, will update the suggested change in this PR

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
taylorrassmannlibertyits: I updated to "Missing all physical address streets" since I wanted to clarify all of them are missing. How do you feel about this?

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
taylorrassmannlibertyits: Yes, yes it would ;)

facilities/src/test/java/gov/va/api/lighthouse/facilities/InternalFacilitiesControllerTest.java
```diff
@@ -133,9 +166,31 @@ void collect_createUpdate() {
...
-        _facility("vha_f1", "FL", "South", 1.2, 3.4, List.of(HealthService.MentalHealthCare));
+        _facility(
+            "vha_f1",
+            "FL",
+            "South",
+            null,
+            null,
+            null,
+            null,
+            null,
```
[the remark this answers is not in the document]
taylorrassmannlibertyits: I had that as a place holder to cycle through a few different missing options. Never truly necessary. Removed and updated.

jblefkove-va (approved):

jnguy-va (approved):

PR #158, opened 2020-09-23
F: Update how missing zip codes are handled

taylorrassmannlibertyits:
Added regex and additional testing for malformed zip codes.


facilities/src/test/java/gov/va/api/lighthouse/facilities/InternalFacilitiesControllerTest.java
```diff
@@ -221,6 +221,7 @@ void collect_noStateOrZip() {
         _facility("vha_f1", "FL", "South", 1.2, 3.4, List.of(HealthService.MentalHealthCare));
     f1.attributes().address().physical().state(null);
     f1.attributes().address().physical().zip(null);
+    f1.attributes().address().mailing(Address.builder().zip("12345-56").build());
```
as-va, line 224: Malformed mailing address is tested, but no test for malformed physical address.

taylorrassmannlibertyits: Since the pattern match is used in both instances I didn't test both cases. Additionally it will introduce a bit more code, but that's what unit tests are for! Update incoming.

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
```diff
@@ -48,6 +49,10 @@
 @AllArgsConstructor(onConstructor = @__(@Autowired))
 @RequestMapping(value = "/internal/management", produces = "application/json")
 public class InternalFacilitiesController {
+  static final String regex = "^[0-9]{5}(?:-[0-9]{4})?$";
```
jblefkove-va, line 52: Does this need to be separate vs. inlining in the pattern definition?

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
```diff
@@ -48,6 +49,10 @@
...
+
+  static final Pattern pattern = Pattern.compile(regex);
```
jblefkove-va, line 54: Should be private if not referenced anywhere else. Name should indicate it's a zip pattern.

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
```diff
@@ -330,6 +335,10 @@ void updateAndSave(ReloadResponse response, FacilityEntity record, Facility faci
       response
           .problems()
           .add(ReloadResponse.Problem.of(facility.id(), "Missing physical address zip"));
+    } else if (!pattern.matcher(record.zip()).matches()) {
```
jblefkove-va, line 338: Would it be sufficient to combine this check with the one above? Either missing or malformed zip could emit `Invalid physical address zip` similar to how invalid latitude/longitude are handled.

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
```diff
@@ -355,6 +364,12 @@ void updateAndSave(ReloadResponse response, FacilityEntity record, Facility faci
         response
             .problems()
             .add(ReloadResponse.Problem.of(facility.id(), "Missing mailing address zip"));
+      } else if (!pattern
+          .matcher(addressMailing(facility).map(a -> a.zip()).toString())
```
jblefkove-va, line 368: `Optional.toString()` is not doing what you expect here

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
jblefkove-va, line 52: Since we're not doing anything fancy with the groups, the `?:` is unnecessary

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
taylorrassmannlibertyits: I was debating that, and I think that should be the way to go, since if the zip is missing it is also invalid.

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
taylorrassmannlibertyits: Would this be more appropriate: `addressMailing(facility).map(a -> a.zip()).map(Object::toString).orElse(null)`

jblefkove-va: Since you only do this check when the zip is not blank, you can safely unwrap the optional to get the underlying string

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
taylorrassmannlibertyits: I ran this through a few different patterns and I think it can all be together.

jnguy-va (approved):

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
```diff
@@ -48,6 +49,10 @@
 @AllArgsConstructor(onConstructor = @__(@Autowired))
 @RequestMapping(value = "/internal/management", produces = "application/json")
 public class InternalFacilitiesController {
+  private static final String zipRegex = "^[0-9]{5}(-[0-9]{4})?$";
+
+  private static final Pattern zipPattern = Pattern.compile(zipRegex);
```
jblefkove-va, line 54: should be `ZIP_PATTERN` since its constant

taylorrassmannlibertyits: First time in years I messed up a constant like that. Thanks for the catch!

jblefkove-va (approved):

PR #159, opened 2020-09-28
F: Reachable Website Notification

taylorrassmannlibertyits:
https://vajira.max.gov/browse/API-2407
Updated code and unit test for invalid website checking during facilities reload.


facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
```diff
@@ -424,6 +428,25 @@ void updateAndSave(ReloadResponse response, FacilityEntity record, Facility faci
...
     }
+
+    if (!isBlank(facility.attributes().website())) {
+
+      HttpClient client = HttpClient.newHttpClient();
+      HttpRequest request =
+          HttpRequest.newBuilder().uri(URI.create(facility.attributes().website())).build();
+
+      HttpResponse<String> responseCode =
+          client.send(request, HttpResponse.BodyHandlers.ofString());
```
jblefkove-va, line 439: Need to handle exceptions and convert into a response problem. Any issues connecting to a URL should not disrupt reloading.

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
```diff
@@ -424,6 +428,25 @@ void updateAndSave(ReloadResponse response, FacilityEntity record, Facility faci
             && isBlank(services(facility).map(s -> s.benefits())))) {
       response.problems().add(ReloadResponse.Problem.of(facility.id(), "Missing services"));
     }
+
+    if (!isBlank(facility.attributes().website())) {
+
+      HttpClient client = HttpClient.newHttpClient();
```
jblefkove-va, line 434: Interesting, does `HttpClient` protect against the things SpotBugs was complaining about for using `HttpURLConnection` directly?

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
```diff
@@ -424,6 +428,25 @@ void updateAndSave(ReloadResponse response, FacilityEntity record, Facility faci
...
+      HttpResponse<String> responseCode =
+          client.send(request, HttpResponse.BodyHandlers.ofString());
+
+      if (responseCode.statusCode() != 200) {
+        response
+            .problems()
+            .add(
+                ReloadResponse.Problem.of(
+                    facility.id(),
+                    "Invalid website provided! Status code: " + responseCode.statusCode()));
```
jblefkove-va, line 447: Since the other problem messages don't have punctuation, I think you could shorten this, e.g. `Invalid website, status code 500`

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
taylorrassmannlibertyits: It does! Talked with Josh about it, and since it uses a builder for the request and doesn't open the connection it resolves the SpotBugs issue.

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
taylorrassmannlibertyits: Ah I agree, I assumed the `SneakyThrows` would take care, but I see that it will just kill the method. What do you think the response problem should look like?

jblefkove-va: I guess `Invalid website` since there won't be a status code

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
```diff
@@ -435,16 +436,19 @@ void updateAndSave(ReloadResponse response, FacilityEntity record, Facility faci
       HttpRequest request =
           HttpRequest.newBuilder().uri(URI.create(facility.attributes().website())).build();
```
jblefkove-va, line 437: Should these lines also be in the `try` block?

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
```diff
@@ -435,16 +436,19 @@ void updateAndSave(ReloadResponse response, FacilityEntity record, Facility faci
...
+            client.send(request, HttpResponse.BodyHandlers.ofString());
+
+        if (responseCode.statusCode() != 200) {
+          response
+              .problems()
+              .add(
+                  ReloadResponse.Problem.of(
+                      facility.id(), "Invalid website, status code: " + responseCode.statusCode()));
+        }
+      } catch (IOException | InterruptedException exception) {
```
jblefkove-va, line 450: Although sometimes frowned on, in this case I think it makes sense to generically catch all `Exception`s. (Or `Throwable` if static analysis complains)

taylorrassmannlibertyits: Ok, that's what I was thinking too. I should have just asked. I did `IOException | InterruptedException` due to what possibly is thrown from `client.send`

jblefkove-va (approved):

PR #160, opened 2020-09-30
Removed 124 Nutrition stop code.

as-va:
https://vajira.max.gov/browse/API-2461


facilities/src/test/java/gov/va/api/lighthouse/facilities/collector/HealthsCollectorJpaTest.java
```diff
@@ -87,9 +87,8 @@ void collect() {
     _initDatabase();
     _saveMentalHealthContact("666", "867-5309", 5555D);
     _saveStopCode("666", "123", "", "10");
```
taylorrassmannlibertyits, line 89: does save stop code require 4 params or can it be reduced to three?

as-va: The `_saveStopCode` function takes four values which are populated into the test database as a test stop code asset. The third parameter (which I assume is the one you are asking about) is the name field for the service. It is a required field for the test data, but since it has no impact on the behavior of this test function it is not given a meaningful value on setup.

So this parameter *could* be refactored out by just having the `saveStopCode` function set all name fields to `""`, my preference would be to leave alone given that (1) the function as written has more flexibility for other tests to use it that *do* care about the name field and (2) this pattern is in use throughout the tests which would mean either a larger effort to refactor *all* tests or refactoring just this one and having it be inconsistent with the rest of the code.

taylorrassmannlibertyits (approved):

jnguy-va (approved):

PR #165, opened 2020-10-26
Search by VISN

jnguy-va:
https://vajira.max.gov/browse/API-285

Adds a new search by VISN endpoint to facilities.

Example local calls:
`http://localhost:8085/v0/facilities?visn=8`
`http://localhost:8085/v0/facilities?visn=10`

If you have issues with running locally with a SQL error you need to re-instantiate the Docker database.


jblefkove-va:
You may already be working on this, but to add a new database column, make sure you update the Liquibase change log (which is what ultimately sets the DB schema) and InternalFacilitiesController, which has the facilities reload code where the entities ultimately are generated.

jnguy-va:
Not sure if we want to edit the System Definitions, let me know if there's another method that's preferable.

Also forgot I need to do swagger changes as well so that's incoming.

facilities/src/main/resources/db/changelog/db.changelog-master.yaml
```diff
@@ -299,3 +299,17 @@ databaseChangeLog:
...
+      id: 9
+      author: axolotl
+      changes:
+        addColumn:
+          tableName: facility
+          schemaName: app
+          columns:
+            - column:
+                name: VISN
+                type: varchar(8)
```
jnguy-va, line 313: I *think* VISNs are always numbers, but not 100% sure.  If so we can change it to int, else leave it as char/string.

facilities-tests/src/test/java/gov/va/api/lighthouse/facilities/tests/SearchIT.java
```diff
@@ -228,6 +228,14 @@ void searchByStateWithTypeAndServices() {
     facilitiesRequest(null, request, 200).expectValid(FacilitiesResponse.class);
   }
 
+  @Test
+  void searchByVisn() {
+    final String state = systemDefinition().ids().visn();
```
jblefkove-va, line 233: visn

facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilitiesController.java
```diff
@@ -269,6 +277,38 @@ private FacilityEntity entityById(String id) {
     return opt.get();
   }
 
+  /** Get facilities by VISN. */
+  @GetMapping(
+      value = "/facilities",
+      produces = "application/json",
```
jblefkove-va, line 283: For consistency there should probably be a `geo+json` version of the endpoint as well.

facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilityRepository.java
```diff
@@ -113,6 +113,28 @@ public Predicate toPredicate(
     }
   }
 
+  @Value
+  @Builder
+  final class VisnSpecification implements Specification<FacilityEntity> {
```
jblefkove-va, line 118: Since this is such a simple search, the JPA specification is overkill. A `List<FacilityEntity> findByVisn(String visn)` method should suffice.

facilities/src/main/resources/db/changelog/db.changelog-master.yaml
jblefkove-va: Since it ultimately boils down to string matching, keep it as varchar.

facilities/src/main/resources/db/changelog/db.changelog-master.yaml
```diff
@@ -299,3 +299,17 @@ databaseChangeLog:
...
+  - changeSet:
+      id: 9
+      author: axolotl
+      changes:
+        addColumn:
+          tableName: facility
+          schemaName: app
+          columns:
+            - column:
+                name: VISN
```
jblefkove-va, line 312: Not case sensitive, but the other column names are lower case.

facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilitiesController.java
```diff
@@ -239,6 +247,14 @@ String allCsv() {
         PageRequest.of(page - 1, perPage, FacilityEntity.naturalOrder()));
   }
 
+  private Page<FacilityEntity> entitiesPageByVisn(String visn, int page, int perPage) {
```
taylorrassmannlibertyits, line 250: other than the name of this method, what is the difference between this and line 226's method `entitiesByVisn`?

facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilitiesController.java
```diff
@@ -368,6 +416,26 @@ GeoFacilitiesResponse geoFacilitiesByState(
         .build();
   }
 
+  /** Get facilities by VISN. */
```
taylorrassmannlibertyits, line 419: Should this and lines 288 - 319 exist? Seems like they are doing the same thing except with `GeoFacilities`

facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilitiesController.java
```diff
@@ -269,6 +285,38 @@ private FacilityEntity entityById(String id) {
...
+  @GetMapping(
+      value = "/facilities",
+      produces = "application/json",
+      params = {"visn"})
+  FacilitiesResponse facilitiesByVisn(
+      @RequestParam(value = "visn") String visn,
+      @RequestParam(value = "page", defaultValue = "1") @Min(1) int page,
+      @RequestParam(value = "per_page", defaultValue = "10") @Min(0) int perPage) {
```
as-va, line 296: This method defines a `per_page` min of 0, but passes this to a function that restricts `per_page` to <= 1.

facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilitiesController.java
jnguy-va: The return responses are different, one is JSON and the other is GeoJSON.

If we want to put them together then I think we'd want to mirror that change with all the endpoints.  Currently the rest of them are seperated.

facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilitiesController.java
jnguy-va: Resolved; `entitiesPageByVisn` and `entitiesByVisn` removed with change from JPA.

facilities/src/main/resources/db/changelog/db.changelog-master.yaml
jnguy-va: Changed to lowercase.  Requires a new DB standup+reload.

facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilityRepository.java
jnguy-va: Changed to `List<>` from JPA

facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilitiesController.java
jnguy-va: Added in!  `geo+json` as well as `vnd.geo+json`

facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilitiesController.java
jnguy-va: Removed the class that had the `per_page` check due to changeover from JPA

facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilitiesController.java
```diff
@@ -269,6 +269,35 @@ private FacilityEntity entityById(String id) {
...
+    List<FacilityEntity> entitiesPage = facilityRepository.findByVisn(visn);
```
jblefkove-va, line 281: this can be `entities` since it is all the entities, not a page

facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilitiesController.java
```diff
@@ -269,6 +269,35 @@ private FacilityEntity entityById(String id) {
     return opt.get();
   }
 
+  /** Get facilities by VISN. */
+  @GetMapping(
+      value = "/facilities",
+      produces = "application/json",
+      params = {"visn"})
+  FacilitiesResponse facilitiesByVisn(
```
jblefkove-va, line 277: `jsonFacilitiesByVisn` to be consistent with the other method names

taylorrassmannlibertyits:
I'm going over the task in JIRA and I wanted to verify some info:

- Do not allow users to search by VISN and Type=benefit. 
I believe this is true based on the code but would like confirmation?
- A VISN is a collection of VA health facilities, lat/log would not a valid search parameter. 
Just want to confirm we're only searching by VISN and cannot add any other params?
- VISN will be used by internal VA consumers, there is not a reason a search for VISN services would be valuable at this time. 
Just want to confirm we're only searching by VISN and we do not ally any VISN services in search?
- This is only available for health facilities and there will always be a VISN associated with a VA health facility.
I don't know if this one holds true? Is there a way to only say search for health facilities?

taylorrassmannlibertyits:
I see you added a searchable column for VISN in the `yaml` file, however, does this automatically update `facilities-local-db`?

jnguy-va:
> I'm going over the task in JIRA and I wanted to verify some info:
> 
> * Do not allow users to search by VISN and Type=benefit.
>   I believe this is true based on the code but would like confirmation?
It does not explicitly filter for type but I just tried it out as a search parameter and it doesn't throw it.  Do we want to implement an invalid parameter on it or leave it as is?


> * A VISN is a collection of VA health facilities, lat/log would not a valid search parameter.
>   Just want to confirm we're only searching by VISN and cannot add any other params?
Same as above, it won't filter by lat/long but it won't give an invalid parameter in the return.


> * VISN will be used by internal VA consumers, there is not a reason a search for VISN services would be valuable at this time.
>   Just want to confirm we're only searching by VISN and we do not ally any VISN services in search?
My thought was that the `services` came from the `clinical-services` table which does not have VISN.  If this is incorrect then we can explicitly filter for `type=health` from the results and then return that?  Can also put this in as a failsafe.


> * This is only available for health facilities and there will always be a VISN associated with a VA health facility.
>   I don't know if this one holds true? Is there a way to only say search for health facilities?
The only VISN pulled in that I saw was from `VAST` and `mental-health-contact` which to my understanding were all health facilities.   If this is incorrect then we can explicitly filter for `type=health` from the results and then return that?  Can also put this in as a failsafe.  (Like the comment above.)

jblefkove-va:
I think those first two bullet points are cruft; the original version of the story. There was discussion about implementing the same facility-type and services parameters that the other search endpoints support, before deciding to simplify to `VISN` only. (Filter parameters can be added in the future if desired.) Supposedly `VISN` is only populated for health facilities, so the endpoint will coincidentally return only health facilities. (I don't think explicit type filtering to only health facilities is necessary.)

Most of the logic in `facilities-local-db` is for populating the facility-collector database, i.e. the CDW tables. For the facilities operational database, `facilities-local-db` just creates an empty database; the schema there is populated automatically by Liquibase when you first start the app.

as-va:
So does that mean that the endpoint will accept lat/long or type as a parameter, but return results without processing it? Would that not be confusing to a user? Wouldn't it be better to return an InvalidParameter in that case? I'm sure there must be a precedent across API on how this is handled, I just haven't been around long enough to know what it is yet.

as-va:
It looks like other search types use a parameter exclusion syntax to enforce only allowing one search type per request, otherwise it returns a BadRequest. See [here](https://github.com/department-of-veterans-affairs/lighthouse-facilities/blob/master/facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilitiesController.java#L508). I think we should put that here to exclude the other search types from the visn search _and_ add !visn to the other existing search types.

as-va (approved):

PR #166, opened 2020-11-05
API-3211 Documentation Updates for new VISN search parameter

jnguy-va:
https://vajira.max.gov/browse/API-3211

Added swagger documentation for the new search by visn parameter.

I feel like I'm missing a file but I *think* this is all?

After team approval will tag UX for review.


facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/FacilitiesSearchApi.java
```diff
@@ -178,6 +180,13 @@ FacilitiesResponse getFacilitiesByLocation(
               style = ParameterStyle.FORM,
               explode = Explode.TRUE)
           List<String> services,
+      @Parameter(
+              name = "visn",
+              in = ParameterIn.QUERY,
+              description =
+                      "Searches all facilities to return any with the matching VISN.",
```
jblefkove-va, line 187: Other parameters have the format `XXX to search for facilities`

(Also while we're in here, can you make `Facilities` on line 130 lowercase to be consistent with the others.

facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/FacilitiesSearchApi.java
```diff
@@ -178,6 +180,13 @@ FacilitiesResponse getFacilitiesByLocation(
               style = ParameterStyle.FORM,
               explode = Explode.TRUE)
           List<String> services,
+      @Parameter(
+              name = "visn",
```
jblefkove-va, line 184: Can we move this up, say after `bbox[]`? So that it is grouped with the search parameters, vs. `type` and `services[]` filter parameters.

facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/FacilitiesSearchApi.java
```diff
@@ -60,6 +60,8 @@
               + "\n\n"
               + "- `zip`, with the option of any combination of `type` or `services[]`"
               + "\n\n"
+              + "- `visn`"
```
jblefkove-va, line 63: alphabetical

jblefkove-va: update line 26 to include visn there as well

jnguy-va: Moved alphabetical and added `visn` to description.

facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/FacilitiesSearchApi.java
```diff
@@ -162,6 +164,12 @@ FacilitiesResponse getFacilitiesByLocation(
                       maxItems = 4,
                       schema = @Schema(type = "number", format = "float")))
           List<BigDecimal> bbox,
+      @Parameter(
+              name = "visn",
+              in = ParameterIn.QUERY,
+              description = "VISN search of matching facilities.",
+              schema = @Schema(type = "number"))
```
as-va, line 171: Is this right? Can this be alphanumeric?

jnguy-va: From previous projects I know that VISNs can only be numbers.  I'm not 100% sure for facilities but I put number in.
VISNs should only be from 1-12 and 15-23 if we want to limit it further.

One map example:
https://www.va.gov/HOMELESS/Supportive_Services_for_Veteran_Families.asp

as-va: Ok. I was remembering from the db update I thought we had made this column a VARCHAR.

jnguy-va: It is a VARCHAR in the db.
`mental-health-contact.csv` has it as `V04`, `V09`, `V19`, etc. and then we change it to just a number in the output.  `"visn": "8"`.  We remove the V but it stays as a string value within the code.

We can have the documentation either way, matching it as a 'String' to the code or as 'integer' to what the value should be.

as-va: Ok, then we should be good to leave as number since this is describing the input the user supplies to the API and it seems like we only accept the numeric portion of the visn as a parameter.

facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/FacilitiesSearchApi.java
darduda: Let's also capitalize VISN. Is VISN defined elsewhere in the documentation, as Veterans Integrated Service Networks? If not, we should define this in the description.

facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/FacilitiesSearchApi.java
```diff
@@ -125,7 +127,7 @@ FacilitiesResponse getFacilitiesByLocation(
               name = "state",
               in = ParameterIn.QUERY,
               description =
-                  "State to search for Facilities. "
+                  "State to search for facilities. "
                       + "Except in rare cases, this is two characters.",
```
darduda, line 131: Let's say "State in which to search for facilities." Can we also list the cases in which the state isn't two characters? I don't know these off the top of my head and am not sure a developer would either.

facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/FacilitiesSearchApi.java
darduda, line 171: where does the user get the VISN numbers? let's say that somewhere if not already.

facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/FacilitiesSearchApi.java
as-va: If VISN is referenced elsewhere in text, it should be capitalized, but this line is documenting the format of the search string which takes all arguments as lower case so in this _particular_ instance it should stay lower case.

facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/FacilitiesSearchApi.java
as-va: From what we were told, this parameter is not for general use, but is to make it easier for VA to generate internal reports. It seems if someone has need of using the VISN search, they would know what the VISN is. I think we could fluff this description string a bit and define VISN, but we don't need to list all the VISNs here. See `id`: we just say it is the id for a facility, but not how to find out the id for a facility.

facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/FacilitiesSearchApi.java
as-va: This is existing verbiage not from this update. Is this something you can find and update easily since we're already in this file?

as-va (approved):

PR #167, opened 2020-11-23
Api 257 use national xml for cemetery website urls

taylorrassmannlibertyits:
https://vajira.max.gov/browse/API-257
Updated the facilities collector to consume a third data-source, https://www.cem.va.gov/cems/national.xml. This is only used for national cemetery website urls.


taylorrassmannlibertyits:
@jnguy-va Fixed the url problem :) Please re-test / re-review when ready 🎉

jnguy-va (approved):
Responses look good to me now!

PR #168, opened 2020-11-23
API-3465 Update swagger schema documentation to match current implementation

jnguy-va:
https://vajira.max.gov/browse/API-3465

Took a couple of other attempts but I believe the `@JsonPropertyOrder` is the easiest way to reorder the swagger output.  LMK otherwise.


facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/v0/Facility.java
```diff
@@ -134,6 +135,24 @@
...
+    "long",
+    "address",
+    "phone",
+    "hours",
+    "services",
+    "satisfaction",
+    "wait_times",
+    "mobile",
+    "active_status",
+    "operating_status",
```
jblefkove-va, line 153: should the `operatingStatus` field also be moved before `visn`?

jnguy-va:
Before and after:
image

facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/v0/Facility.java
jnguy-va: Looks like it's in order in `GeoFacility` but not `Facility`

taylorrassmannlibertyits (approved):

PR #180, opened 2020-12-31
API-000 Adding exception logging

jnguy-va:
Adding logging to debug daily `reload` error.

Slack thread: https://libertyits.slack.com/archives/CBVUWQ21Y/p1609254589309000


as-va (approved):

as-va (approved):

jblefkove-va:
This PR should also restore `<skipPush>true</skipPush>` to the facilities-tests pom so that deployments can run.

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
```diff
@@ -277,6 +277,7 @@ private void moveToGraveyard(ReloadResponse response, FacilityEntity entity) {
         processMissingFacility(response, missingId);
       }
     } catch (Exception e) {
+      log.error("Failed to process missing facilities: {}", e.getMessage());
```
jblefkove-va, line 280: `Failed to process facilities` is probably more appropriate; the exception may also have occurred in `updateFacility()`

jblefkove-va (approved):

PR #184, opened 2021-01-06
Api 4217 covid vaccine service filtering

taylorrassmannlibertyits:
https://vajira.max.gov/browse/API-4217


jnguy-va (approved):
Testing after merge in QA required but looks good!

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
```diff
@@ -229,14 +227,6 @@ GraveyardResponse graveyardAll() {
                                                 MAPPER,
                                                 z.cmsOperatingStatus(),
                                                 Facility.OperatingStatus.class))
-                                    .cmsServices(
```
as-va, line 232: Do we not need `cmsServices` here?

facilities/src/main/java/gov/va/api/lighthouse/facilities/CmsOverlayController.java
```diff
@@ -54,8 +57,19 @@ void initDirectFieldAccess(DataBinder dataBinder) {
...
 
     if (overlay.cmsServices() != null) {
-      entity.cmsServices(
-          FacilitiesJacksonConfig.createMapper().writeValueAsString(overlay.cmsServices()));
+      // Compile the set of overlay services to be added
+      Set<String> detailedServices = new HashSet<>();
+      for (Facility.CmsService service : overlay.cmsServices()) {
+        // Since the covid 19 service name doesn't match our enum, we need to update and verify??
+        if (1 == service.active()) {
+          if (service.name().equals("COVID-19 vaccines")) {
```
as-va, line 65: I don't like having this transform logic hardcoded and buried in the code. This makes maintenance more difficult as we continue to add new services.

I'd rather have a `Map<String, String>` with service names as keys and HealthService as values. This map would contain all the mappings and be a single point of updating as new features are added. Then instead of having to maintain this function every time a new service is added with an ever growing set of `if` blocks, this code then could simply be:
```
if (1 == service.active()) {
    detailedServices.add(serviceMap.getOrDefault(service.name(), service.name())
}
```

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
taylorrassmannlibertyits: We do not since we are no longer storing this.

facilities/src/main/java/gov/va/api/lighthouse/facilities/CmsOverlayController.java
taylorrassmannlibertyits: I really like this idea. _However_ in the interest of time, could we make this a new task for the next sprint? We also don't know if we are only going to receive HealthService data from CMS? That would be something that needs verification.

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
taylorrassmannlibertyits: But we SHOULD store `overlayServices` updating now

as-va: Yes, _something_ needed to be saved. Looks ok now.

facilities/src/main/java/gov/va/api/lighthouse/facilities/CmsOverlayController.java
as-va: In the interest of closing out the Sprint and getting the feature operational quickly, will add this as a future improvement.

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
```diff
@@ -169,10 +169,8 @@ static String zipOf(Facility facility) {
       log.info("Facility {} does not exist, ignoring request.", sanitize(id));
       return ResponseEntity.accepted().build();
     }
-    if (entity.get().cmsOperatingStatus() != null || entity.get().cmsServices() != null) {
-      log.info(
-          "Failed to delete facility {}. cmsOperatingStatus or cmsServices are not null.",
-          sanitize(id));
+    if (entity.get().cmsOperatingStatus() != null || entity.get().overlayServices() != null) {
+      log.info("Failed to delete facility {}. cmsOperatingStatus", sanitize(id));
```
as-va, line 173: What happened with this line? This is no longer a complete thought.

as-va (approved):

PR #185, opened 2021-01-06
API-4254 Handle missing latitude and longitude in reload

jnguy-va:
https://vajira.max.gov/browse/API-4254

Added a catch for missing lat/longs coming from NCA.
This can be tested by editing an entry in the `nca.csv` file to have missing lat/long and then building and hitting the `reload` endpoint.

There's various places this check can be put in.  If anyone thinks another spot is better I'm open to suggestions.


facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
```diff
@@ -85,8 +85,12 @@ private static boolean isHoursNull(Facility facility) {
   @SneakyThrows
   static FacilityEntity populate(FacilityEntity record, Facility facility) {
     checkArgument(record.id() != null);
-    record.latitude(facility.attributes().latitude().doubleValue());
-    record.longitude(facility.attributes().longitude().doubleValue());
+    try {
+      record.latitude(facility.attributes().latitude().doubleValue());
+      record.longitude(facility.attributes().longitude().doubleValue());
+    } catch (IllegalArgumentException e) {
```
jblefkove-va, line 91: This is not the right kind of exception to catch. (`IllegalArgumentException` vs `NullPointerException`)

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
```diff
@@ -85,8 +85,12 @@ private static boolean isHoursNull(Facility facility) {
   @SneakyThrows
   static FacilityEntity populate(FacilityEntity record, Facility facility) {
     checkArgument(record.id() != null);
-    record.latitude(facility.attributes().latitude().doubleValue());
-    record.longitude(facility.attributes().longitude().doubleValue());
+    try {
+      record.latitude(facility.attributes().latitude().doubleValue());
```
jblefkove-va, line 89: Rather than catching the exception, how about a null-check that defaults to zero, e.g:

`record.latitude(Optional.ofNullable(facility.attributes().latitude()).orElse(BigDecimal.ZERO).doubleValue());`

as-va: That seems like something we probably shouldn't just decide to do out of the blue. That would have to be agreed to in an ICD or other contract between us and consumers of the API.

Also, that would set each field independently of the other although lat/long is really just _one_ location, so we could end up in a situation where we modify one field but the other is valid so we've just "transported" a facility to somewhere on the equator or meridian.

taylorrassmannlibertyits: I suggest we put an invalid value here instead of 0 aka any value that that is outside the range of spatial coordinates since we catch this error further down in the code.

facilities/src/test/java/gov/va/api/lighthouse/facilities/InternalFacilitiesControllerTest.java
```diff
@@ -216,7 +216,8 @@ void collect_missing() {
     when(collector.collectFacilities()).thenReturn(List.of(f1));
     ReloadResponse response = _controller().reload().getBody();
     assertThat(response.facilitiesUpdated()).isEqualTo(List.of("vha_f1"));
-    assertThat(response.facilitiesMissing()).isEqualTo(List.of("vha_f2", "vha_f3", "vha_f4"));
+    assertThat(response.facilitiesMissing())
```
taylorrassmannlibertyits, line 219: did this just change indentation?

taylorrassmannlibertyits: *More carriage return than indentation ;)

jnguy-va: kinda... line too long so formatter split it in half

jblefkove-va: Run `mvn process-sources` and commit again

as-va (approved):

taylorrassmannlibertyits (approved):

PR #188, opened 2021-01-14
Moved ReloadResponse to facilities-api and updated tests for coverage.

taylorrassmannlibertyits:


facilities-api/src/test/java/gov/va/api/lighthouse/facilities/api/v0/ReloadResponseTest.java
```diff
@@ -0,0 +1,18 @@
+package gov.va.api.lighthouse.facilities.api.v0;
+
+import org.junit.jupiter.api.Test;
+
+public class ReloadResponseTest {
+
+  // There are much more thorough tests for ReloadResponse, however, for coverage this test was
+  // required.
+  // See ReloadIT.java or InternalFacilitiesControllerTest.java for more
```
jnguy-va, line 9: Does this not work if you put it inside of `InternalFacilitiesControllerTest.java`?

jnguy-va (dismissed):
Quick blanket tests passed, deferring to @as-va  for final approval.

facilities-api/src/test/java/gov/va/api/lighthouse/facilities/api/v0/ReloadResponseTest.java
taylorrassmannlibertyits: Unfortunately no. There are calls to that area of code in the `InternalFacilitiesControllerTest` already and the coverage wasn't picked up. I tested your suggestion to make sure too and nope :'(

as-va: The coverage isn't picked up in the other module because only the module being tested gets instrumented and code coverage can only be captured from the instrumented version of the file, therefore you can't (*) have a test in one module generate code coverage on a source file in a different module.

(*) Well, you _can_ configure maven/jacoco to do this, but I don't thinks that's the direction we should go.

facilities-api/src/test/java/gov/va/api/lighthouse/facilities/api/v0/ReloadResponseTest.java
```diff
@@ -0,0 +1,18 @@
...
+  @Test
+  void reloadCoverageTest() {
```
as-va, line 11: This is not a real test. The file path, file name, class name, function annotation, and function name all say it is, but they are all lying.

`InternalFacilitiesControllerTest` may be covering some of this (`ReloadResponse`) class by nature of using it, but this class is now owned by the `facilities-api` module, and each module is responsible for testing the code within it. I think we should have a real set of tests here.

as-va (changes requested):

facilities-api/src/test/java/gov/va/api/lighthouse/facilities/api/v0/ReloadResponseTest.java
taylorrassmannlibertyits: This creates a large dependency change to the entire project to fully test ReloadResponse in this new module. Is your decision for me to spend time and effort to migrate this over since it is already completed in the other module?

as-va: Lets talk offline about scope and LOE concerns.

facilities-api/src/test/java/gov/va/api/lighthouse/facilities/api/v0/ReloadResponseTest.java
jblefkove-va: It should be straightforward to turn these into real tests, no? i.e. add an assertion to check that `markComplete()` set the expected timestamp?

taylorrassmannlibertyits: I was overcomplicating the issue. Will make a simple test for this. Thanks all.

as-va (approved):

PR #190, opened 2021-01-18
API-4205 Adding check for missing VISN field on reload

jnguy-va:
https://vajira.max.gov/browse/API-4205

Check for if VISN is blank.

Wondering if we should check if VISN is valid?  There are only specific numbers that can be a VISN:
**1 2 4 5 6 7 8 9 10 12 15 16 17 19 20 21 22 23**    [1, 2, 4-10, 12 and 15–23]


facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
```diff
@@ -439,6 +439,9 @@ void updateAndSave(ReloadResponse response, FacilityEntity record, Facility faci
             && isBlank(services(facility).map(s -> s.benefits())))) {
       response.problems().add(ReloadResponse.Problem.of(facility.id(), "Missing services"));
     }
+    if (isBlank(record.visn())) {
```
taylorrassmannlibertyits, line 442: I don't think this is specific enough. We should specify the facility type that associates with VISN and only do this check for those. Otherwise we would have a lot of missing VISN values.

jnguy-va: Only `va_health_facility` and `vet_center`s have VISNs associated with them.  Benefits and cemetery do not.

taylorrassmannlibertyits: https://vajira.max.gov/browse/API-4205?focusedCommentId=789075&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-789075

taylorrassmannlibertyits (dismissed):

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
```diff
@@ -439,6 +439,12 @@ void updateAndSave(ReloadResponse response, FacilityEntity record, Facility faci
             && isBlank(services(facility).map(s -> s.benefits())))) {
       response.problems().add(ReloadResponse.Problem.of(facility.id(), "Missing services"));
     }
+    if ((facility.attributes().facilityType() == Facility.FacilityType.va_health_facility
```
as-va, line 442: While this works, it is a bit verbose. ```((A && C) || (B && C))``` is equivalent to ```((A || B) && C)```, so this can become
```
if ((facility.attributes().facilityType() == Facility.FacilityType.va_health_facility || 
    facility.attributes().facilityType() == Facility.FacilityType.vet_center) && 
    isBlank(record.visn()))
```
or even
```
if (Arrays.asList(Facility.FacilityType.va_health_facility, Facility.FacilityType.vet_center)
    .contains(facility.attributes().facilityType() && isBlank(record.visn()))
```

facilities/src/test/java/gov/va/api/lighthouse/facilities/InternalFacilitiesControllerTest.java
```diff
@@ -296,6 +298,60 @@ void collect_noStateOrZip() {
                 ReloadResponse.Problem.of("vha_f1", "Invalid longitude")));
   }
 
+  @Test
+  @SneakyThrows
+  void collect_noVISNHealth() {
+    Facility f1 = _facility("vha_f1", "FL", "32934", 91.4, 181.4, List.of());
+    f1.attributes().facilityType(va_health_facility);
```
as-va, line 305: These two tests are exactly the same except for the facility type being tested, so you could make these one test that loops to do each type without needing to duplicate so much code.

You could replace this line with
```
for (Facility.FacilityType facilityType : Arrays.asList(va_health_facility, vet_center)) {
    f1.attributes().facilityType(facilityType)
    ...
}
````

File: facilities/src/test/java/gov/va/api/lighthouse/facilities/InternalFacilitiesControllerTest.java [hunk 4]
```diff
@@ -296,6 +298,60 @@ void collect_noStateOrZip() {
...
+    when(collector.collectFacilities()).thenReturn(List.of(f1));
+    ReloadResponse response = _controller().reload().getBody();
+    assertThat(response.facilitiesCreated()).isEqualTo(List.of("vha_f1"));
+    assertThat(response.problems())
```
as-va, line 309: This test is only testing the behavior of a facility with no VISN, so that's really the only `Problem` response we need to check for. Other `Problem` types should be tested elsewhere. Checking every field in every test like this will make updating the tests a chore every time anything changes in the class.

You can just check that the "Missing VISN" `Problem` is in the list of returned `Problem`s.

facilities/src/test/java/gov/va/api/lighthouse/facilities/InternalFacilitiesControllerTest.java
jnguy-va: Tried the loop method, but it requires another input as the enums for `id` is required to match.  (_vha_ and _vc_)

as-va (dismissed):

as-va (approved):

PR #199, opened 2021-02-22
Api 5023 national xml as cem primary name source quick fix update

taylorrassmannlibertyits:


facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/CemeteriesCollector.java
```diff
@@ -152,6 +152,11 @@ static CdwCemetery toCdwCemetery(ResultSet rs) {
   private String xmlFacilityName(
       List<NationalCemeteries.NationalCemetery> cemeteries, String siteId) {
     for (NationalCemeteries.NationalCemetery cem : cemeteries) {
+
+      if (cem.id.length() == 2) {
```
as-va, line 156: Only three digit ids are valid and we've already seen cases where there are fewer digits (hence this bugfix). Might as well make this catch potential single digit cases too.

as-va, line 156: Seems we would never want to use a two digit id as only a three digit representation is valid for our use. Instead of copying this transformation every place the id is used, would be better to place where the XML is read so it will always be correct for the rest of the code and we won't find this bug again when someone attempts to use without copying this code snippet.

as-va (changes requested):
Cleanup of fix.

as-va (approved):

PR #201, opened 2021-03-11
API-5211 Fortify fixes for Facilities 2021

jnguy-va:
**DO NOT DELETE THIS BRANCH AFTER MERGE**

Will need it as reference until Fortify has been all-clear from the SWA team.

.
.
.
https://vajira.max.gov/browse/API-5211

### DIFF
--- a/facilities/src/main/java/gov/va/api/lighthouse/facilities/BingResponse.java
+++ b/facilities/src/main/java/gov/va/api/lighthouse/facilities/BingResponse.java
@@ -29,7 +29,7 @@ public final class BingResponse {
   @Builder
   @JsonIgnoreProperties(ignoreUnknown = true)
   static final class Resource {
-    Point point;
+    Point resourcePoint;
 
     @JsonIgnoreProperties(ignoreUnknown = true)
     static final class ResourceBuilder {}
--- a/facilities/src/main/java/gov/va/api/lighthouse/facilities/BingResponse.java
+++ b/facilities/src/main/java/gov/va/api/lighthouse/facilities/BingResponse.java
@@ -1,6 +1,7 @@
 package gov.va.api.lighthouse.facilities;
 
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
 import java.math.BigDecimal;
 import java.util.ArrayList;
 import java.util.List;
@@ -29,6 +30,7 @@ public final class BingResponse {
   @Builder
   @JsonIgnoreProperties(ignoreUnknown = true)
   static final class Resource {
+    @JsonProperty("point")
     Point resourcePoint;
 
     @JsonIgnoreProperties(ignoreUnknown = true)
--- a/facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilityEntity.java
+++ b/facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilityEntity.java
@@ -187,7 +187,7 @@ public class FacilityEntity implements HasFacilityPayload {
     static Optional<Pk> optionalFromIdString(@NonNull String typeAndStationNumber) {
       try {
         return Optional.ofNullable(fromIdString(typeAndStationNumber));
-      } catch (Exception ex) {
+      } catch (NullPointerException ex) {
         return Optional.empty();
       }
     }
--- a/facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilityEntity.java
+++ b/facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilityEntity.java
@@ -187,7 +187,7 @@ public class FacilityEntity implements HasFacilityPayload {
     static Optional<Pk> optionalFromIdString(@NonNull String typeAndStationNumber) {
       try {
         return Optional.ofNullable(fromIdString(typeAndStationNumber));
-      } catch (NullPointerException ex) {
+      } catch (IllegalArgumentException ex) {
         return Optional.empty();
       }
     }
--- a/facilities/src/main/java/gov/va/api/lighthouse/facilities/HealthController.java
+++ b/facilities/src/main/java/gov/va/api/lighthouse/facilities/HealthController.java
@@ -108,14 +108,9 @@ public class HealthController {
       statusCode = response.getStatusCode();
       JsonNode root = JacksonConfig.createMapper().readTree(response.getBody());
       checkState(!((ArrayNode) root).isEmpty(), "No %s entries", name);
-    } catch (RestClientException e) {
-      log.info("Rest client exception occurred. GET {} message: {}", url, e.getMessage());
-      statusCode = HttpStatus.SERVICE_UNAVAILABLE;
-    } catch (JsonProcessingException e) {
-      log.info("JSON processing exception occurred. GET {} message: {}", url, e.getMessage());
-      statusCode = HttpStatus.SERVICE_UNAVAILABLE;
-    } catch (IllegalArgumentException e) {
-      log.info("Illegal argument exception occurred. GET {} message: {}", url, e.getMessage());
+    } catch (RestClientException | JsonProcessingException | IllegalArgumentException e) {
+      log.info(
+          "{} occurred. GET {} message: {}", e.getClass().getSimpleName(), url, e.getMessage());
       statusCode = HttpStatus.SERVICE_UNAVAILABLE;
     }
     return Health.status(new Status(statusCode.is2xxSuccessful() ? "UP" : "DOWN", name))
--- a/facilities/src/main/java/gov/va/api/lighthouse/facilities/HealthController.java
+++ b/facilities/src/main/java/gov/va/api/lighthouse/facilities/HealthController.java
@@ -6,6 +6,8 @@ import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.node.ArrayNode;
 import gov.va.api.health.autoconfig.configuration.JacksonConfig;
 import gov.va.api.lighthouse.facilities.collector.InsecureRestTemplateProvider;
+
+import java.io.IOException;
 import java.sql.Timestamp;
 import java.time.Instant;
 import java.time.temporal.ChronoUnit;
@@ -106,7 +108,7 @@ public class HealthController {
       statusCode = response.getStatusCode();
       JsonNode root = JacksonConfig.createMapper().readTree(response.getBody());
       checkState(!((ArrayNode) root).isEmpty(), "No %s entries", name);
-    } catch (Exception e) {
+    } catch (IOException e) {
       log.info("Exception occurred. GET {} message: {}", url, e.getMessage());
       statusCode = HttpStatus.SERVICE_UNAVAILABLE;
     }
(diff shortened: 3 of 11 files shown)


facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilityEntity.java
```diff
@@ -187,7 +187,7 @@ static Pk fromIdString(@NonNull String typeAndStationNumber) {
     static Optional<Pk> optionalFromIdString(@NonNull String typeAndStationNumber) {
       try {
         return Optional.ofNullable(fromIdString(typeAndStationNumber));
-      } catch (Exception ex) {
+      } catch (NullPointerException ex) {
```
as-va, line 190: Shouldn't be able to get a NPE in this line because this function itself has already verified the parameter is not null before passing to the next function in the chain. Maybe the analyzer tool is not smart enough to realize this and only notices that the called function has a `NonNull` annotation. Shouldn't hurt to have this in any event.

jnguy-va:
Looks like these exceptions were needed to cover all the test bases and requirements of `readTree`.  This caused coverage to drop... looking for places to add coverage.  Having some trouble trying to mock out the JsonProcessingException.

facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilityEntity.java
jnguy-va: Changed this to IllegalArgumentException, which is what the unit tests were throwing.

facilities/src/main/java/gov/va/api/lighthouse/facilities/HealthController.java
```diff
@@ -106,7 +108,7 @@ private static Health testHealthJsonList(
       statusCode = response.getStatusCode();
       JsonNode root = JacksonConfig.createMapper().readTree(response.getBody());
       checkState(!((ArrayNode) root).isEmpty(), "No %s entries", name);
-    } catch (Exception e) {
+    } catch (IOException e) {
```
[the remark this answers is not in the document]
jnguy-va: Added `IllegalArgumentException` and `RestClientException` to cover the old unit tests and added `JsonProcessingException` as a requirement for `readTree`.

facilities/src/main/java/gov/va/api/lighthouse/facilities/Parameters.java
```diff
@@ -41,10 +41,12 @@ Parameters add(String key, @NonNull Object value) {
   }
 
   Parameters addAll(String key, List<?> values) {
-    if (!isEmpty(values)) {
-      for (Object val : values) {
-        checkArgument(val != null);
-        params.add(key, val.toString());
+    if (values != null) {
```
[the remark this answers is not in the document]
jnguy-va: Note to self: Need to verify if Fortify is okay with this.

facilities/src/main/java/gov/va/api/lighthouse/facilities/HealthController.java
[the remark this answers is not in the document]
as-va: Can anything in this block actually throw an `IllegalArgumentException` or did the test just need to raise any exception type and picked that one randomly to use? If the later, we shouldn't add production code in to catch an exception that only exists in a unit test, instead the unit test should be changed to test with an actual exception type that can be raised in this case.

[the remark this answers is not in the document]
jnguy-va: This unit test: `HealthControllerTest.collectorBackendHealth_oneOrMoreUnhealthyBackendServicesReturns503AndHealthWithDownStatuses` throws an `IllegalArgumentException` due to the content being null for ResponseEntity<Health>.  I'm guessing this could happen in production when an endpoint is active but the body content is null?

facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilityEntity.java
jnguy-va: If the values for determining Facility's ID is invalid it will be caught in this exception.  The unit tests put in a malformed string `x`, `xxx` etc.  Are there any other instances besides unmatchable fields to determine Facility ID and null that could occur?

facilities/src/main/java/gov/va/api/lighthouse/facilities/HealthController.java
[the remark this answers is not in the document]
as-va: Ah, ok. I did not see the `IllegalArgumentException` in the call path of anything in this block but going down a couple extra levels I see it undocumented in the `readtree` path. So yes, IAE should stay.

facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilityEntity.java
as-va: Yes, this is correct then.

facilities/src/main/java/gov/va/api/lighthouse/facilities/HealthController.java
```diff
@@ -106,8 +108,14 @@ private static Health testHealthJsonList(
       statusCode = response.getStatusCode();
       JsonNode root = JacksonConfig.createMapper().readTree(response.getBody());
       checkState(!((ArrayNode) root).isEmpty(), "No %s entries", name);
-    } catch (Exception e) {
-      log.info("Exception occurred. GET {} message: {}", url, e.getMessage());
+    } catch (RestClientException e) {
```
as-va, line 111: These catch blocks all have the same body except for the string including their name. So this can be simplified from a bunch of separate blocks to one:

    } catch (RestClientException | JsonProcessingException | IllegalArgumentException e) {
        log.info("{} occurred. GET {} message: {}", e.getClass().getSimpleName(), url, e.getMessage());

as-va: Side note: this build is failing because of code coverage violation. You must be right on the very edge of the cut off since there is very little new code added in this PR. Getting rid of these unnecessary `catch`es may solve the code coverage problem at the same time.

taylorrassmannlibertyits: Local builds should also be able to catch this coverage failure.

facilities/src/test/java/gov/va/api/lighthouse/facilities/HealthControllerTest.java
```diff
@@ -9,6 +9,7 @@
 import static org.mockito.Mockito.when;
 import static org.springframework.web.context.request.RequestContextHolder.setRequestAttributes;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
```
taylorrassmannlibertyits, line 12: I think this can be removed since nothing else is modified in the class. `mvn clean install` _should_ do this for you.

facilities/src/test/java/gov/va/api/lighthouse/facilities/HealthControllerTest.java
```diff
@@ -77,6 +86,7 @@ private HealthController _controller() {
 
   @Test
   void collectionStatusHealth_clearCache() {
+    // //
```
taylorrassmannlibertyits, line 89: please remove commented out comment

facilities/src/test/java/gov/va/api/lighthouse/facilities/HealthControllerTest.java
```diff
@@ -123,6 +133,7 @@ void collectorBackendHealth_allHealthyBackendServicesReturns200AndHealthWithUpSt
         .thenReturn(ok());
     when(jdbcTemplate.queryForObject(any(String.class), eq(Timestamp.class)))
         .thenReturn(Timestamp.from(Instant.now()));
+    // /////
```
taylorrassmannlibertyits, line 136: please remove commented out comment

facilities/src/test/java/gov/va/api/lighthouse/facilities/HealthControllerTest.java
```diff
@@ -137,6 +148,7 @@ void collectorBackendHealth_allHealthyBackendServicesReturns200AndHealthWithUpSt
 
   @Test
   void collectorBackendHealth_clearCache() {
+    // //
```
taylorrassmannlibertyits, line 151: please remove commented out comment

facilities/src/test/java/gov/va/api/lighthouse/facilities/HealthControllerTest.java
```diff
@@ -160,6 +172,7 @@ void collectorBackendHealth_clearCache() {
         .thenReturn(ok());
     when(jdbcTemplate.queryForObject(any(String.class), eq(Timestamp.class)))
         .thenReturn(Timestamp.from(Instant.now().minus(48, ChronoUnit.HOURS)));
+    // ////
```
taylorrassmannlibertyits, line 175: please remove commented out comment

facilities/src/test/java/gov/va/api/lighthouse/facilities/HealthControllerTest.java
```diff
@@ -203,6 +214,7 @@ void collectorBackendHealth_servicesUnreachable() {
         .thenReturn(ok());
     when(jdbcTemplate.queryForObject(any(String.class), eq(Timestamp.class)))
         .thenReturn(Timestamp.from(Instant.now().minus(48, ChronoUnit.HOURS)));
+    // ////////
```
taylorrassmannlibertyits, line 217: please remove commented out comment

facilities/src/test/java/gov/va/api/lighthouse/facilities/HealthControllerTest.java
```diff
@@ -64,6 +64,16 @@ private static void assertStatus(Health overallStatus, ExpectedStatus expected)
     return new ResponseEntity<>("[{}]", HttpStatus.OK);
   }
 
+  private HealthController _badController() {
```
taylorrassmannlibertyits, line 67: Is this used anywhere in the test?

jnguy-va: Not anymore :)

facilities/src/main/java/gov/va/api/lighthouse/facilities/BingResponse.java
```diff
@@ -29,7 +30,8 @@
   @Builder
   @JsonIgnoreProperties(ignoreUnknown = true)
   static final class Resource {
-    Point point;
+    @JsonProperty("point")
```
taylorrassmannlibertyits, line 33: @jnguy-va Since this is a builder from a specific response (Bing's own api) when you changed `point` to `resourcePoint` you needed to have this `JsonProperty`.

taylorrassmannlibertyits (approved):

facilities/src/main/java/gov/va/api/lighthouse/facilities/BingResponse.java
jnguy-va: =O

as-va: This class is built from the response we get back from the Bing API call. The response has a field called `point` which is what this class is supposed to hold. By default, the Builder framework maps field from the data to an instance of the object by matching the name of the data field to the name of the class member field. Changing this name made the names no longer match so the framework failed to figure out how to parse the response. the `@JsonProperty` annotation overwrites this behavior by saying that we want this field to hold what is listed in that markup line `"point"` instead of matching on the field name `"resourcePoint"`.

PR #202, opened 2021-03-24
Update CMS Overlay Path with URL from CSV file

taylorrassmannlibertyits:
https://vajira.max.gov/browse/API-5645
Updated to replace path from cms overlay with csv path


facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/FacilitiesCollector.java
```diff
@@ -135,7 +136,7 @@ static String withTrailingSlash(@NonNull String url) {
     Map<String, String> websites;
     Collection<VastEntity> vastEntities;
     try {
-      websites = loadWebsites();
+      websites = loadWebsites("websites.csv");
```
jnguy-va, line 139: I know Fortify caught it on CCE but not on here for some reason, but can we set `websites.csv` as a global variable?  It might come up as a forced injection warning.

facilities/src/main/java/gov/va/api/lighthouse/facilities/CmsOverlayController.java
```diff
@@ -60,6 +62,13 @@ void initDirectFieldAccess(DataBinder dataBinder) {
       // Compile the set of overlay services to be added
       Set<String> detailedServices = new HashSet<>();
       for (DetailedService service : overlay.detailedServices()) {
+
+        if (service.name().equals("COVID-19 vaccines")) {
+          Map<String, String> websites = loadWebsites("COVID-19-Facility-URLs.csv");
```
jnguy-va, line 67: I know Fortify caught it on CCE but not on here for some reason, but can we set `COVID-19-Facility-URLs.csv` as a global variable?  It might come up as a forced injection warning.

jnguy-va: On second though do you think these should live in a properties file or nah?

facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/FacilitiesCollector.java
jnguy-va: On second though do you think these should live in a properties file or nah?

taylorrassmannlibertyits: I think making it a constant up at the top should suffice.

facilities/src/main/java/gov/va/api/lighthouse/facilities/CmsOverlayController.java
taylorrassmannlibertyits: I think making it a constant up at the top should suffice.

facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/cms/DetailedService.java
```diff
@@ -74,7 +74,6 @@
   int onlineSchedulingAvailable;
 
   // todo example?
```
as-va, line 76: Leftover comment.

facilities/src/main/java/gov/va/api/lighthouse/facilities/CmsOverlayController.java
```diff
@@ -60,6 +64,13 @@ void initDirectFieldAccess(DataBinder dataBinder) {
       // Compile the set of overlay services to be added
       Set<String> detailedServices = new HashSet<>();
       for (DetailedService service : overlay.detailedServices()) {
+
+        if (service.name().equals("COVID-19 vaccines")) {
```
as-va, line 68: Can this be combined with the other identical `if` check below? The difference is the other is only executed when the service is active. So the question then becomes: do we populate inactive detailed services entries in the Response if they are inactive? I think the answer should be `no`, in which case this would be safe to move below as we won't care what the value is if the service is inactive.

as-va:
I do not see any test files updated as part of this change. There should be a test added to verify the path in a Response object matches the value from the csv and not the incoming overlay.

facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/cms/DetailedService.java
taylorrassmannlibertyits: No, I left this because we don't have concrete examples from CMS yet. This is also in the 5058 draft PR.

facilities/src/main/java/gov/va/api/lighthouse/facilities/CmsOverlayController.java
taylorrassmannlibertyits: Ok sounds good. I should have just asked you, since I originally had it that way! :)

taylorrassmannlibertyits:
> I do not see any test files updated as part of this change. There should be a test added to verify the path in a Response object matches the value from the csv and not the incoming overlay.

Will update in next push

facilities/src/test/java/gov/va/api/lighthouse/facilities/CmsOverlayControllerTest.java
```diff
@@ -50,8 +50,7 @@ private CmsOverlay overlay() {
                     .appointmentLeadIn(
                         "Your VA health care team will contact you if you...more text")
                     .onlineSchedulingAvailable(0)
-                    .path(
-                        "\\/erie-health-care\\/locations\\/erie-va-medical-center\\/covid-19-vaccines")
+                    .path("https://www.maine.va.gov/services/covid-19-vaccines.asp")
```
as-va, line 53: This doesn't seem like it actually tests what we want to verify. We should test saving an overlay with a "garbage" path, read the object back out, and verify the reported path is equal to the url csv entry for the given facility id rather than the input path.

taylorrassmannlibertyits: Ah I see, will update

as-va (approved):

PR #203, opened 2021-03-25
Api 5238 update drive time band entity for monthyear field

taylorrassmannlibertyits:


facilities/src/main/resources/db/changelog/db.changelog-master.yaml
```diff
@@ -444,3 +444,14 @@ databaseChangeLog:
...
+      id: 15
+      author: axolotl
+      changes:
+        addColumn:
+          tableName: drive_time_band
+          schemaName: app
+          columns:
+            - column:
+                name: month_year
+                type: varchar(16)
```
as-va, line 457: This is bigger than needed. All month_year values are of the form `XXX####`, so the max (and min) string size we'd need to store is `7`.

as-va (approved):

PR #207, opened 2021-03-31
LOAST Test Result Changes for Facilities

jnguy-va:
https://vajira.max.gov/browse/API-5933


facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/FacilitiesSearchApi.java
```diff
@@ -146,7 +146,7 @@ FacilitiesResponse getFacilitiesByLocation(
                   "Latitude of point to search for facilities, "
                       + "in WGS84 coordinate reference system.",
               schema = @Schema(type = "number", format = "float"),
-              examples = @ExampleObject(name = "coordinates", value = "40.0"))
+              examples = @ExampleObject(name = "coordinates", value = "123.4"))
```
taylorrassmannlibertyits, line 149: This is an invalid latitude range

taylorrassmannlibertyits (dismissed):

facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/v0/GeoFacility.java
```diff
@@ -124,7 +126,7 @@
     @JsonProperty(value = "detailed_services")
```
taylorrassmannlibertyits, line 126: Make sure to add nullable true here as well (already added in Facility.java)

jnguy-va:
Looking at this...

facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/v0/NearbyResponse.java
```diff
@@ -17,7 +17,8 @@
 
   public enum Type {
     @JsonProperty("nearby_facility")
-    NearbyFacility
+    NearbyFacility,
+    nearby_facility
```
taylorrassmannlibertyits, line 21: Why was this added? This means there is a property type we were missing before?

jnguy-va: Removed this change with the latest push, but the LOAST testing tool is flagging that "nearby_facility" isn't a valid enum for Type.  Trying to make a work around, but so far I think it might be something on the test tool end that needs to support @JsonProperty within enums.

taylorrassmannlibertyits: In that case I think you should remove this new enum value. Especially since it is unused.

taylorrassmannlibertyits (dismissed):

jnguy-va:
From the CMS Documentation PR: https://github.com/department-of-veterans-affairs/lighthouse-facilities/pull/208#issuecomment-817774002

> A quick note on the discussion of periods in schema descriptions (for consistency) - there is no internal consistency within the Facilities API with regard to periods appearing at the end of schema description lines. In fact, the vast majority do not have terminating periods prior to this update. May want to check for a company/program style guide prior to making sweeping changes.

I see the periods have been added to that PR, do we want them to also be cleaned up for the rest of Facilities here?

facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/v0/NearbyResponse.java
jnguy-va: Sent some feedback to the LOAST team, waiting for reply if we need to fix or if it's on their end not supporting the JSON property name.

facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/v0/Facility.java
```diff
@@ -197,26 +202,31 @@
...
-    @Schema(description = "This field is deprecated and replaced with \"operating_status\".")
+    @Schema(
+        description = "This field is deprecated and replaced with \"operating_status\".",
+        nullable = true)
     ActiveStatus activeStatus;
 
     @Valid
     @NotNull
     @JsonProperty(value = "operating_status", required = true)
+    @Schema(example = "A")
```
as-va, line 221: This is not a valid example for `operating_status` - this is an example of an `active_status`. See the `OperatingStatus` class for a description the `operating_status` structure.

jnguy-va:
There is currently a swagger bug with serializing enums with @JsonProperty or @JsonValue.

https://github.com/swagger-api/swagger-core/pull/3553
https://github.com/openapi-tools/swagger-maven-plugin/issues/52

as-va (approved):

PR #209, opened 2021-04-09
F: PSSG drive-time band update consumer notification 

taylorrassmannlibertyits:
https://vajira.max.gov/browse/API-5930
Updated nearby endpoint for lat / lon search to include 'band_version' in it's output


facilities/src/main/java/gov/va/api/lighthouse/facilities/NearbyController.java
```diff
@@ -263,8 +263,15 @@ NearbyResponse nearbyLatLong(
       @RequestParam(value = "drive_time", required = false) Integer maxDriveTime) {
     List<NearbyId> ids = nearbyIds(longitude, latitude, services, maxDriveTime);
 
+    // Extract the month year out of the first NearbyId
+    String monthYear = "Unknown";
+    if (!ids.isEmpty() && driveTimeBandRepository.findById(ids.get(0).bandId).isPresent()) {
+      monthYear = driveTimeBandRepository.findById(ids.get(0).bandId).get().monthYear();
```
as-va, line 269: This leads to an odd implication. 
* If we find a list of valid ids, this will populate the version field as expected.
* If we find a list of ids but the bands are corrupted (which should have been caught in the band update process), this will return the bands with an "Unknown" version, which is fine
* HOWEVER, if the nearby search results in no ids matching the search criteria, this will return an empty facility list (correct) with a band version "Unknown" (odd)

We _do_ know the band version - we should be able to tell the user which version of bands were used that resulted in a report of no nearby facilities.

taylorrassmannlibertyits: I slightly disagree here. We know band versions of _specific_ bands. We do not know a "generic" band version. Maybe the alternative is "No bands found for versioning"?

as-va: We do know it - it's in the database (thousands of times) - you just don't have it readily available here. We can have a discussion on what we might be able to do with the db to get this out in a general case.

facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/v0/NearbyResponse.java
```diff
@@ -39,6 +39,16 @@
     Integer maxTime;
   }
 
+  @Value
+  @Builder
+  @JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY)
+  @Schema(description = "JSON API-compliant object containing metadata about a nearby VA facility")
```
as-va, line 45: Instead of `a nearby VA facility`, I think this should be"
* `nearby VA facilities`
* `a nearby facilities response`
* `this response`
* `the data in this response`
* etc
I just don't like the singular reference to a facility as this API returns a list of multiple facilities.

taylorrassmannlibertyits: Sounds good, I will make it `this response` since I don't want it to only be about facilities. Good catch.

as-va (approved):

as-va:
Can you go ahead and add this to the address endpoint as well so we can close this ticket?

taylorrassmannlibertyits:
> Can you go ahead and add this to the address endpoint as well so we can close this ticket?

Sure, I was working a higher priority item, but I'll add this shortly

as-va (approved):

PR #214, opened 2021-05-07
F: Create telehealth endpoint

taylorrassmannlibertyits:
https://vajira.max.gov/browse/API-6872
Updated facilities to have new telehealth get and post eps.


facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/FacilitiesTelehealthApi.java
```diff
@@ -0,0 +1,64 @@
...
+              mediaType = "application/json",
+              schema = @Schema(implementation = GenericError.class)))
+  @ApiResponse(
+      responseCode = "403",
+      description = "Invalid API token",
+      content =
+          @Content(
+              mediaType = "application/json",
+              schema = @Schema(implementation = GenericError.class)))
+  @ApiResponse(
```
as-va, line 50: Do we need a response block for `429`s? See 6878. Since these are being worked in tandem, that task won't add an update to this branch so it should be added here if needed.

facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/v0/TelehealthBody.java
```diff
@@ -0,0 +1,19 @@
...
+import io.swagger.v3.oas.annotations.media.Schema;
+import lombok.Builder;
+import lombok.Data;
+
+@Data
+@Builder
+@JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY)
+@Schema(
+    description =
+        "Data provided by telehealth team to Facilities to be applied on top of known data.")
```
as-va, line 14: I know this is just a stub, but the telehealth endpoint only returns the same data that it receives - it is not 'applied on top of known data'.

facilities-tests/src/test/java/gov/va/api/lighthouse/facilities/tests/TelehealthIT.java
```diff
@@ -0,0 +1,55 @@
...
+                .contentType("application/json")
+                .body(MAPPER.writeValueAsString(TelehealthBody.builder().stub("stub").build()))
+                .request(Method.POST, svc.urlWithApiPath() + "v0/telehealth/" + id + "/update"))
+        .expect(200);
+  }
+
+  @Test
+  void telehealthById() {
+    final String id = systemDefinition().ids().facility();
+    final String request = "v0/telehealth?id=" + id;
```
as-va, line 52: This shouldn't work? The post path in the controller is (currently) `v0/telehealth/{id}/update`

facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilitiesController.java
```diff
@@ -665,6 +671,27 @@ FacilityReadResponse readJson(@PathVariable("id") String id) {
     return FacilityReadResponse.builder().facility(facility(entityById(id))).build();
   }
 
+  @GetMapping(
```
as-va, line 674: More food for thought: should the telehealth implementation be part of the existing `FacilitiesController` or should we create a new `TelehealthController`?

facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/FacilitiesTelehealthApi.java
taylorrassmannlibertyits: Agreed, if we do keep this documentation I'll add the 429

facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/v0/TelehealthBody.java
taylorrassmannlibertyits: I slightly disagree here. This is for POST'ing. We are going to only overwrite fields supplied in this body (I think). Not blank out anything that is not provided in the body?

facilities-tests/src/test/java/gov/va/api/lighthouse/facilities/tests/TelehealthIT.java
taylorrassmannlibertyits: This is the `GET` not the `POST` test.

facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilitiesController.java
taylorrassmannlibertyits: I am thinking the same thing. I'm not 100% sold on putting this in the `FacilitiesController`. While the direction in the ticket was to place in here, (from a previously worked ticket on endpoint creation) I think this is more similar to the `CmsController` and should be by itself?

facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/v0/TelehealthBody.java
as-va: Ah, I read 'known data' as info we already know from `/facilities` - that is what I was arguing against. I don't believe VCM is far enough along to commit to how/if they will do partial POSTs, but from the screenshots and descriptions of the tools it seems likely we would get the full payload every time. But that isn't worth pressing on just to comment a stub.

jnguy-va: Perhaps "...on top of known telehealth data."?

facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilitiesController.java
as-va: Looking at the rest of the project structure, I do think breaking it out into its own controller class would be best.

facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/FacilitiesTelehealthApi.java
taylorrassmannlibertyits: Added `429` in latest push

jnguy-va (approved):

facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/FacilitiesTelehealthApi.java
```diff
@@ -0,0 +1,64 @@
...
+import io.swagger.v3.oas.annotations.enums.ParameterIn;
+import io.swagger.v3.oas.annotations.media.Content;
+import io.swagger.v3.oas.annotations.media.ExampleObject;
+import io.swagger.v3.oas.annotations.media.Schema;
+import io.swagger.v3.oas.annotations.responses.ApiResponse;
+import io.swagger.v3.oas.annotations.security.SecurityRequirement;
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+
+public interface FacilitiesTelehealthApi {
```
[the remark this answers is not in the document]
as-va: Status update: still awaiting a response on this before we can close it out.

PR #216, opened 2021-05-17
F: Update Operational_Hours_Special_Instructions Payload

taylorrassmannlibertyits:
https://vajira.max.gov/browse/API-7117
Updated to find and replace special instruction strings with testing and coverage


facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
```diff
@@ -51,6 +51,28 @@
 @AllArgsConstructor(onConstructor = @__(@Autowired))
 @RequestMapping(value = "/internal/management", produces = "application/json")
 public class InternalFacilitiesController {
+  static final String SPECIAL_INSTRUCTION_ONE =
```
as-va, line 54: Recommend naming these `SPECIAL_INSTRUCTION_OLD_1` and `SPECIAL_INSTRUCTION_[NEW/CHANGED/UPDATED]_1`. I feel like the format here for the original doesn't carry enough context for a future user to understand its purpose easily enough. Plus, the alphabetical ordering is nonsensical with the current pattern. Finally, spelling out the numbers is needlessly verbose; if this expands, who wants to end up typing SPECIAL_INSTRUCTION_UPDATE_TWENTY_FOUR?

as-va: Or you could just store these all as a map of `{old : new}` mappings and not have to number the entries at all with the additional benefit of simplifying the replacement function, but that would be a bit more of an architectural change than renaming.

facilities/src/test/java/gov/va/api/lighthouse/facilities/InternalFacilitiesControllerTest.java
```diff
@@ -645,6 +639,39 @@ void updateAndSave_error() {
...
+    Facility f3 = _facility("vha_f3", "FL", "32934", 5.6, 6.7, List.of(HealthService.UrgentCare));
+    f1.attributes().operationalHoursSpecialInstructions(SPECIAL_INSTRUCTION_ONE);
+    f2.attributes().operationalHoursSpecialInstructions(SPECIAL_INSTRUCTION_TWO);
+    f3.attributes().operationalHoursSpecialInstructions(SPECIAL_INSTRUCTION_THREE);
+    ReloadResponse response = ReloadResponse.start();
+    controller.updateAndSave(
+        response,
+        FacilityEntity.builder().id(FacilityEntity.Pk.fromIdString("vha_f1")).build(),
+        f1);
+    assertThat(f1.attributes().operationalHoursSpecialInstructions())
```
as-va, line 659: This verifies that the original object is modified, but not that it was modified _prior_ to being saved. A better test would pass these test objects to the `updateAndSave` function then read them back from the repo and verify the field on the stored objects is the modified version.

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
taylorrassmannlibertyits: I'll stick with the simple (non-map) way for now since we have so few. If the list expands more in the future we can change it up.

as-va (approved):

PR #221, opened 2021-06-07
API-7315 TimeZoneMap and Tests

jnguy-va:
https://vajira.max.gov/browse/API-7315

Maven gave me a bit of a heartache but it works now :)

Imported the https://github.com/dustin-johnson/timezonemap library used to calculate Olsen time from a Latitude/Longitude.
Maven link: https://mvnrepository.com/artifact/us.dustinj.timezonemap/timezonemap/4.5

Note: Branched off the Java 16 PR.


facilities/src/test/java/gov/va/api/lighthouse/facilities/TimeZoneMapTest.java
```diff
@@ -0,0 +1,86 @@
+package gov.va.api.lighthouse.facilities;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import org.junit.jupiter.api.Test;
+import us.dustinj.timezonemap.TimeZoneMap;
+
+public class TimeZoneMapTest {
+  public TimeZoneMap calculateTimeZoneMap(double latitude, double longitude) {
+    return TimeZoneMap.forRegion(latitude - 1, longitude - 1, latitude + 1, longitude + 1);
```
taylorrassmannlibertyits, line 10: Why is there a `+/-1` for everything?

facilities/pom.xml
```diff
@@ -55,6 +57,12 @@
       <artifactId>commons-csv</artifactId>
       <version>${commons-csv.version}</version>
     </dependency>
+    <!-- https://mvnrepository.com/artifact/us.dustinj.timezonemap/timezonemap -->
```
taylorrassmannlibertyits, line 60: Do we need to download this independently ?

jnguy-va: Maven should see it and download it when you build Facilities.

facilities/src/test/java/gov/va/api/lighthouse/facilities/TimeZoneMapTest.java
jnguy-va: What this library does is load a sector of the map and then does it's calculations on it, specifically any long/lat you provide within the boundaries.

It's a pretty time-greedy calculation for figuring out the region, and gets exponentially longer the larger the area.  (I let it try to load the entire continental US and it didn't finish in 5 minutes.)

So to make it on-demand, just load the small snippet zone around the lat/tlong which is less then a second for calculating it.

facilities/pom.xml
jnguy-va: Not sure if we want this comment in here, but I put it in for future reference =/

taylorrassmannlibertyits: I like the comment in there

facilities/src/test/java/gov/va/api/lighthouse/facilities/TimeZoneMapTest.java
taylorrassmannlibertyits: Oh I think I better understand here. So will the `+/-1` be part of how we calculate the regular timezone? I think that's a pretty large area if we're talking real longitude and latitude values.

jnguy-va: Yes, it will be used to calculate the timezone.  I'm pretty sure we can make it a smaller value though, like +/- 0.01 or 0.1 etc.

taylorrassmannlibertyits: Cool. I think we should have the test do what we're going to do IRL. Plus it will make it faster ;)

taylorrassmannlibertyits (dismissed):

facilities/pom.xml
```diff
@@ -17,6 +17,7 @@
     <guava.version>29.0-jre</guava.version>
     <mssql-jdbc.version>7.4.1.jre8</mssql-jdbc.version>
     <swagger-maven-plugin.version>2.1.1</swagger-maven-plugin.version>
+    <timezonemap.version>4.5</timezonemap.version>
```
as-va, line 20: We should be trying to use whatever the latest version is, not just this specific version. Time zones are occasionally updated and we want to automatically get any updates as they are published.

facilities/src/test/java/gov/va/api/lighthouse/facilities/TimeZoneMapTest.java
as-va: I actually think this is _not_ what we should do in the "IRL". When we are processing a reload, we'll be calculating time zones for thousands of points, so we should just take the hit and initialize the map once rather than reinitialize it with a small zone prior to every calculation.

This is fine for this test as it still proves the library works and the tests should be made fast as they are run all the time, but this conversation veered into design of the application code so I'm addressing _that_ only.

jnguy-va: I'll have to see which calculation takes longer, initializing the entire continental US or point-by-point.  My initial gut feeling is that doing every point will be faster then loading the US, but we can figure that out on the next ticket with implementation (unless you want it in this PR instead!).
Aside from Alaska/Hawaii, do we have any Facilities that are way-out-there?

taylorrassmannlibertyits: We do have facilities in US territories, which are in random locations around the world.

as-va: Not just US territories, there are facilities in other countries scattered across the globe. Not a lot, but enough to account for.

taylorrassmannlibertyits (approved):

facilities/src/test/java/gov/va/api/lighthouse/facilities/TimeZoneMapTest.java
```diff
@@ -0,0 +1,90 @@
+package gov.va.api.lighthouse.facilities;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import org.junit.jupiter.api.Test;
+import us.dustinj.timezonemap.TimeZoneMap;
+
+public class TimeZoneMapTest {
+  public TimeZoneMap calculateTimeZoneMap(double latitude, double longitude) {
+    double regionBoundary = 0.0001;
```
taylorrassmannlibertyits, line 10: This can be final

as-va (approved):

facilities/src/test/java/gov/va/api/lighthouse/facilities/TimeZoneMapTest.java
jnguy-va: Will add in the implementation branch

PR #225, opened 2021-06-17
API-7654 Add time zone to facilities

jnguy-va:
https://vajira.max.gov/browse/API-7654

Added in time zones calculations and display for all facilities.


facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/CalculateTimeZone.java
```diff
@@ -0,0 +1,53 @@
...
+          "Time zone calculation failed, latitude [{}] or longitude [{}] is invalid.",
+          latitudeDecimal,
+          longitudeDecimal);
+    }
+    return timeZone;
+  }
+
+  /** Loads the EVERYWHERE map. */
+  @SneakyThrows
+  public static TimeZoneMap everywhereMap() {
```
taylorrassmannlibertyits, line 47: Since this is the map for _everywhere_ I think think this method may not be needed anymore (unless you want to time things). If you do keep the method, you should change the variable return to `TimeZoneMap.forEverywhere()` since it is no longer the usMap

facilities/src/test/java/gov/va/api/lighthouse/facilities/collector/TimeZoneMapTest.java
```diff
@@ -0,0 +1,82 @@
...
+@Slf4j
+public class TimeZoneMapTest {
+  @Test
+  void invalidLatitudeLongitude() {
+    TimeZoneMap usMap = TimeZoneMap.forEverywhere();
+    String invalidLatitudeZone =
+        CalculateTimeZone.calculateTimeZonesWithMap(null, BigDecimal.valueOf(149.915261), usMap);
+    String invalidLongitudeZone =
+        CalculateTimeZone.calculateTimeZonesWithMap(BigDecimal.valueOf(36.737343), null, usMap);
+    assertThat(invalidLatitudeZone).isEqualTo(null);
```
taylorrassmannlibertyits, line 20: You can actually call `.isNull()` here instead of `isEqualTo`

taylorrassmannlibertyits: (This can be done multiple places)

facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/CalculateTimeZone.java
jnguy-va: Did we want to keep this for the future CAPs work?  Or just know that the CAPs interface will have to initialize the map (or possibly not use the EVERYWHERE map but just lat/long?)

taylorrassmannlibertyits: Maybe rename to `getEverywhereTimezoneMap`? Just so it's a little more clear what this method call is doing.

taylorrassmannlibertyits (dismissed):

taylorrassmannlibertyits:
I don't know how to unapprove, but I don't think this is working yet @as-va

PR #226, opened 2021-06-20
Api 7654 add timezone to api response

taylorrassmannlibertyits:
The bulk of this PR work was done by @jnguy-va. Since she is no longer on the team, I am making this "clean up" PR. It was easier (for me) to decline the old PR and create this new branch.


facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/CalculateTimeZone.java
```diff
@@ -0,0 +1,59 @@
...
+import us.dustinj.timezonemap.TimeZoneMap;
+
+/**
+ * The TimeZoneMap library works by: 1. Generating an 'area' from the SW to NE to calculate on, via
+ * lat/long points (TimeZoneMap) 2. Figuring out the time zone from a lat/long point within the map
+ * (TimeZone) 3. Translating this point to Olsen time (getZoneId) Alternatively the EVERYWHERE map
+ * can be used, though it does take longer to initiate.
+ */
+@Slf4j
+public class CalculateTimeZone {
```
as-va, line 18: I don't like this class name. Starting with a verb, it reads like a function name (see first function in class). Let's make this something better. At the very least, invert the construction to `TimeZoneCalculator`. Would also be fine with `TimeZoneFinder`, `TimeZoneUtils`, etc.

as-va: This _works_, but is inefficient and not a good practice in encapsulation. If you make this a Singleton class, you can still have the FacilitiesCollector create the object and take the hit of generating the map, then just have the Transformer classes fetch the instance and make the calculation calls. This:
1. Avoids awkwardly passing the map object all the way through the stack
2. Removes the dependency on `us.dustinj.timezonemap` from the rest of the project and contains it _only_ to here
3. Leaves the `*Collector`s unmodified
4. Eliminates the need for the (very) ugly `@SuppressWarnings("ObjectToString")`s all over the place

as-va (changes requested):

facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/CalculateTimeZone.java
taylorrassmannlibertyits: Updated to be a `@UtilityClass`

facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/TimeZoneFinder.java
```diff
@@ -15,7 +14,10 @@
...
  */
 @Slf4j
-public class CalculateTimeZone {
+@UtilityClass
+public class TimeZoneFinder {
+  public static final TimeZoneMap WORLD_MAP = TimeZoneMap.forEverywhere();
+
   /** Calculate and load timezones given longitude and latitude passing in a map. */
   @SneakyThrows
   public static String calculateTimeZonesWithMap(
```
as-va, line 23: This is only ever used with the `WORLD_MAP`, so there's no need to have every caller pass it in. This can come out of the function definition and just use `WORLD_MAP` directly from this function.

taylorrassmannlibertyits: I was thinking that too, but left it in just in case. I'll remove and update the other classes.

facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/TimeZoneFinder.java
```diff
@@ -0,0 +1,50 @@
...
+import us.dustinj.timezonemap.TimeZoneMap;
+
+/**
+ * The TimeZoneMap library works by: 1. Generating an 'area' from the SW to NE to calculate on, via
+ * lat/long points (TimeZoneMap) 2. Figuring out the time zone from a lat/long point within the map
+ * (TimeZone) 3. Translating this point to Olsen time (getZoneId) Alternatively the EVERYWHERE map
+ * can be used, though it does take longer to initiate.
+ */
+@Slf4j
+@UtilityClass
```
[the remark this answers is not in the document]
taylorrassmannlibertyits: Through my testing, compile times were _faster_ than the previous PR times by 1 minute or greater. This indicates the WORLD_MAP is not loaded each call

as-va (approved):

PR #227, opened 2021-07-01
Initial update for version one upgrade

taylorrassmannlibertyits:
https://vajira.max.gov/browse/API-8058
~~and~~
~~https://vajira.max.gov/browse/API-8059~~

Big update implementing v1 initial classes for upgrade. There is A LOT of duplication here, however, this is intentional. As the team progresses forward with v1 changes for Facilities, the v1 code will diverge from v0.


as-va:
Looks like this is moving forward with bifurcating at the db level with the versioned Facility strings. Did you look into the unversioned Facility db representation and translating after reading like we had discussed last week? Did you identify any problems with the other approach or just think that this is better?

taylorrassmannlibertyits:
> Looks like this is moving forward with bifurcating at the db level with the versioned Facility strings. Did you look into the unversioned Facility db representation and translating after reading like we had discussed last week? Did you identify any problems with the other approach or just think that this is better?

Saving to the db has to be _some version_ and that would be represented as v0 (at the moment). I thought about it a bit more, and saw the implemented path forward as the way to go. I think a daily save of two different facility types is better than having to transform for each call.

xenkidu-va (approved):
LGTM. Looks like a functioning, good starting point.

facilities/src/main/resources/db/changelog/db.changelog-master.yaml
```diff
@@ -455,3 +455,30 @@ databaseChangeLog:
...
+        addColumn:
+          tableName: facility
+          schemaName: app
+          columns:
+            - column:
+                name: facility_v1
+                type: clob
+                constraints:
+                  nullable: true
+  - changeSet:
```
as-va, line 472: Shouldn't need two changesets in one update.

facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/FacilityPair.java
```diff
@@ -0,0 +1,12 @@
+package gov.va.api.lighthouse.facilities.api;
+
+import gov.va.api.lighthouse.facilities.api.v0.Facility;
+import lombok.Builder;
+import lombok.Data;
+
+@Data
+@Builder
+public class FacilityPair {
+  Facility v0;
```
as-va, line 10: In files that need to reference both versions, we should use fully qualified names for _both_ instead of importing one and qualifying the other.

taylorrassmannlibertyits: Ok, I was debating this. My train of thought was all v0 `Facility` objects didn't have the fully qualified name.

facilities/src/main/resources/db/changelog/db.changelog-master.yaml
taylorrassmannlibertyits: In the past when modifying different tables we've made two changeSets.

as-va: I see that happened once in the initial vaccine service update, but it did not need to be that way. I think they got added at different times during that task and they were never coalesced back to a single change entry.

taylorrassmannlibertyits: I'm still uneasy about changing this. I feel there was a reason to keep them separate before.

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
```diff
@@ -79,35 +80,44 @@
 
   private static final ObjectMapper MAPPER = FacilitiesJacksonConfig.createMapper();
```
as-va, line 81: Anything that has to split out a V1 variant should change the original to `*V0`.

facilities/src/main/java/gov/va/api/lighthouse/facilities/ControllersV1.java
```diff
@@ -0,0 +1,92 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.TreeMap;
+import java.util.function.Function;
+import lombok.NonNull;
+import lombok.experimental.UtilityClass;
+
+@UtilityClass
+final class ControllersV1 {
```
as-va, line 29: In many places throughout the code new V1 classes have been created by cloning the existing files and appending "V1", but the old files are unchanged. Either both classes should keep a versionless name and be moved to a versioned package, or the existing implementations should be renamed to `*V0`.

facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/FacilitiesCollector.java
```diff
@@ -225,9 +268,25 @@ static String withTrailingSlash(@NonNull String url) {
...
+            .sorted((left, right) -> left.id().compareToIgnoreCase(right.id()))
+            .collect(toList());
+
+    List<gov.va.api.lighthouse.facilities.api.v1.Facility> facilitiesV1 =
+        Streams.stream(
+                Iterables.concat(benefitsV1, cemeteriesV1, healthsV1, stateCemsV1, vetCentersV1))
+            .sorted((left, right) -> left.id().compareToIgnoreCase(right.id()))
+            .collect(toList());
+
+    for (int i = 0; i < facilities.size(); i++) {
```
as-va, line 284: This makes the implicit assumption that both lists will be the same size and have the same elements in the same order. Only the last of those assumptions seems safe to make.

This should be the case in the _ideal_ situation, but I can see the possibility as v0 and v1 continue to diverge that some facility may meet one version's constraints but not the others causing a different set of Facilities collected and then this logic would break.

A safer approach if this is the way we want to handle collection would be to collect facilities as a map of `{id : obj}` for each version, iterate over the union of the set of ids from both, and extract facility objects from the maps based on id to ensure the pairs are initialized correctly.

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalCollectorController.java
```diff
@@ -51,7 +51,7 @@
   }
 
   @GetMapping(value = "/facilities")
-  List<Facility> collectFacilities() {
+  List<FacilityPair> collectFacilities() {
```
as-va, line 54: This is changing the return type of an exposed endpoint. Are you sure we can make this change? It is an internal endpoint so no consumers can be using it, but could there be any other effects or implications of this change? Could Prometheus, pingdom, or any other monitoring tool be using this endpoint which would break with this change or are we certain it is only a back door for us?

as-va (changes requested):

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
taylorrassmannlibertyits: Ok, I did not do that originally to reduce the amount of files changed. I will make the update.

facilities/src/main/java/gov/va/api/lighthouse/facilities/ControllersV1.java
taylorrassmannlibertyits: Ok, I did not do that originally to reduce the amount of files changed. I will make the update.

as-va: I agree it is a hassle, but without it gives the impression that (e.g.) `Facility` is the _real_ thing and `FacilityV1` is a variant or sub-type instead of them being parallel objects at the same level.

facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/FacilitiesCollector.java
taylorrassmannlibertyits: I disagree. If we have a moment where there are `n` v0 facilities and `m` v1 facilities then something is _fundamentally_ wrong with the API.

Since these are two _sorted_ lists of v0 and v1 Facilities this _should_ be ok. 

I can add a check for if the different lists are not the same size and error out. 

I need more explanation on a scenario you think these two versions may diverge such that the pair option does not work.

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalCollectorController.java
taylorrassmannlibertyits: Since this endpoint is _internal_ I think this change is ok. I will verify and update.

facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/FacilitiesCollector.java
as-va: Yes, there would be something critically wrong, but simply erroring out would put us in the same predicament we had several months ago where were were unable to reload or modify results in the API which seems a bad way to handle an error. It would be better to insert a null for the missing version string, or drop the unmatched object, or _anything_ that would let the API continue to operate with at least the known good data while we work to fix whatever the underlying error is.

taylorrassmannlibertyits: Ok I like that way of thinking. I'll work on a check if the id's match and throw a null into the v0 or v1 pair. I don't thinking dropping the facility would be ok

as-va: null would be better if it works, but I just threw that out without looking back through the full stack. Would that cause a problem with serialization/deserialization later on trying to map the null to a FacilityVx object? I wasn't sure so I added the "drop the unmatched" option out as a backup.

taylorrassmannlibertyits: After looking into this more. I _still_ think this will not happen. The process using the **same** collection for v0 and v1, the transform is the only difference into different versions of facilities.

facilities/src/main/java/gov/va/api/lighthouse/facilities/ControllersV1.java
```diff
@@ -0,0 +1,88 @@
...
+
+  static Set<ServiceType> validateServices(Collection<String> services) {
+    if (isEmpty(services)) {
+      return emptySet();
+    }
+    List<ServiceType> results = new ArrayList<>(services.size());
+    for (String service : services) {
+      ServiceType mapped = SERVICE_LOOKUP.get(trimToEmpty(service));
+      if (mapped == null) {
+        throw new ExceptionsV0.InvalidParameter("services", service);
```
as-va, line 82: Do we need to break this out to have a `V1` version or can this be genericized to a common non-versioned `Exceptions` class?

as-va:
I know for a while you were keeping the documentation changes active so you could compare; since we are getting close to merging this, has that been deactivated? What will this merge do to the portal?

facilities/src/main/resources/db/changelog/db.changelog-master.yaml
as-va: To close this out, this part of the code change has been removed from this PR, but we should revisit this conversation before the follow up work to re-add.

facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/FacilitiesCollector.java
as-va: Recapping an offline conversation: the collection is _currently_ the same, but this may change during the course of V1's lifetime. I think we should still have safe code for inserting into the db, but the db portion of this PR is being pushed back to allow the other pieces to merge so this can be revisited in a follow up task to re-add the db portion.

facilities/src/main/java/gov/va/api/lighthouse/facilities/ControllersV1.java
taylorrassmannlibertyits: Probably can be made into an exceptions utils class

taylorrassmannlibertyits:
> I know for a while you were keeping the documentation changes active so you could compare; since we are getting close to merging this, has that been deactivated? What will this merge do to the portal?

The goal is to hide anything related to v1. This should be disabled in the upcoming push

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalCollectorController.java
taylorrassmannlibertyits: I think we should be good here. I don't see anything else using this

as-va (approved):

PR #232, opened 2021-08-04
Api 8854 v1 cms overlay updates

taylorrassmannlibertyits:
https://vajira.max.gov/browse/API-8954

Updated Facilities API to dynamically update a facility on receipt of CMS Overlay information in regards to upcoming v1 changes. Ripple effect throughout the code base.


facilities/src/main/java/gov/va/api/lighthouse/facilities/GeoFacilityTransformerV1.java
```diff
@@ -1,67 +1,59 @@
...
-    if (isBlank(id())) {
-      return null;
-    }
-    return GeoFacility.builder()
-        .type(GeoFacility.Type.Feature)
-        .geometry(geometry())
-        .properties(properties())
-        .build();
-  }
+  //  @NonNull private final Facility facility;
```
esubocz-va, line 5: is this code going to be used at a later time or can this be deleted?

taylorrassmannlibertyits: Definitely going to be used later. This code should have been commented out from a previous PR since it is not currently in use.

facilities/src/main/java/gov/va/api/lighthouse/facilities/CmsOverlayEntity.java
```diff
@@ -0,0 +1,36 @@
...
+
+@Data
+@Entity
+@Builder
+@Table(name = "cms_overlay", schema = "app")
+@NoArgsConstructor(access = AccessLevel.PUBLIC)
+@EqualsAndHashCode(onlyExplicitlyIncluded = true)
+@AllArgsConstructor(access = AccessLevel.PRIVATE)
+public class CmsOverlayEntity {
+  @EqualsAndHashCode.Include @EmbeddedId private FacilityEntity.Pk id;
```
esubocz-va, line 25: I think this should probably be a non-nullable field.

facilities/src/main/java/gov/va/api/lighthouse/facilities/CmsOverlayController.java
```diff
@@ -50,54 +87,125 @@ void initDirectFieldAccess(DataBinder dataBinder) {
   @SneakyThrows
   ResponseEntity<Void> saveOverlay(
       @PathVariable("id") String id, @Valid @RequestBody CmsOverlay overlay) {
-    Optional<FacilityEntity> existingEntity =
-        repository.findById(FacilityEntity.Pk.fromIdString(id));
-    if (existingEntity.isEmpty()) {
-      log.info("Received Unknown Facility ID ({}) for CMS Overlay", sanitize(id));
-      return ResponseEntity.accepted().build();
+    Optional<FacilityEntity> existingFacilityEntity =
```
esubocz-va, line 90: This method is kind of clunky & tough to follow, could we break it up into some more easily digestible parts?

taylorrassmannlibertyits: Cool, I'll split it up into two separate methods for facility / cms overlay updates.

facilities/src/main/java/gov/va/api/lighthouse/facilities/CmsOverlayController.java
```diff
@@ -50,54 +87,125 @@ void initDirectFieldAccess(DataBinder dataBinder) {
...
+              .id(FacilityEntity.Pk.fromIdString(id))
+              .cmsOperatingStatus(MAPPER_V0.writeValueAsString(overlay.operatingStatus()))
+              .cmsServices(
+                  activeServices.isEmpty() ? null : MAPPER_V0.writeValueAsString(activeServices))
+              .build();
 
-        // Since the covid 19 service name doesn't match our enum, we need to update and verify??
-        if (service.active()) {
+      if (facility != null) {
+        facility.attributes().operatingStatus(overlay.operatingStatus());
```
esubocz-va, line 136: We run the risk of overwriting existing non-null data with null data here.

taylorrassmannlibertyits: Good catch. We had this problem in the past. I'll make sure to make a test since I don't think one exists.

facilities/src/main/java/gov/va/api/lighthouse/facilities/CmsOverlayController.java
```diff
@@ -34,9 +38,57 @@
...
+    }
+    return activeServices;
+  }
+
+  @GetMapping(
+      value = {"/v0/facilities/{id}/cms-overlay", "/v1/facilities/{id}/cms-overlay"},
+      produces = "application/json")
+  @SneakyThrows
+  CmsOverlayResponse getOverlay(@PathVariable("id") String id) {
+    if (cmsOverlayRepository.findById(FacilityEntity.Pk.fromIdString(id)).isPresent()) {
```
as-va, line 69: You could find the entity once then check if it is present to avoid two lookup operations.

facilities/src/main/java/gov/va/api/lighthouse/facilities/CmsOverlayController.java
```diff
@@ -50,54 +102,107 @@ void initDirectFieldAccess(DataBinder dataBinder) {
...
-      return ResponseEntity.accepted().build();
+    Optional<FacilityEntity> existingFacilityEntity =
+        facilityRepository.findById(FacilityEntity.Pk.fromIdString(id));
+    Optional<CmsOverlayEntity> existingCmsOverlayEntity =
+        cmsOverlayRepository.findById(FacilityEntity.Pk.fromIdString(id));
+
+    updateFacilityData(existingFacilityEntity, id, overlay);
+    updateCmsOverlayData(existingCmsOverlayEntity, id, overlay);
+
+    return ResponseEntity.ok().build();
```
as-va, line 113: This changes the existing behavior of the endpoint. I will verify with CMS if this will have any impact on their end.

facilities/src/main/java/gov/va/api/lighthouse/facilities/CmsOverlayController.java
```diff
@@ -50,54 +102,107 @@ void initDirectFieldAccess(DataBinder dataBinder) {
...
+    } else {
+      cmsOverlayEntity = existingCmsOverlayEntity.get();
+
+      if (overlay.operatingStatus() != null) {
+        cmsOverlayEntity.cmsOperatingStatus(
+            MAPPER_V0.writeValueAsString(overlay.operatingStatus()));
+      }
+
+      if (overlay.detailedServices() != null) {
+        cmsOverlayEntity.cmsServices(
```
as-va, line 142: This is incorrect. Maybe there was a miscommunication in discussing this. CMS will always send the full payload for any service that has changed (meaning we do not need to persist any services that transition to inactive since we will get everything back if it is reenabled), but will _not_ send unchanged services. So this needs to:
* Add or replace any service that has `active: true`
* Delete any service that has `active: false`
* Keep all other stored services

as-va (changes requested):

facilities/src/main/java/gov/va/api/lighthouse/facilities/CmsOverlayController.java
taylorrassmannlibertyits: I think this code is correctly doing what you are suggesting. Could you present a case where this does not work?

My logic is:

- Get all active services from the overlay
- Only write those services to the respective `cmsServices` column
- If there was a change from active -> inactive for a service, it is no longer part of the active list so it will be "deleted"

as-va: But this does not preserve any _existing_ services that were saved from previous calls.

If a Facility has services `A`, `B`, and `C` stored in the DB, then CMS turns off service 'B', we will get an overlay with service `B` with `active: false`. We need to remove that service, but keep `A` and `C` in the DB since they are still active.

taylorrassmannlibertyits: Does this mean the "full payload" is only for a changed service, and not the entire list of detailed services? If so, that was communicated to me incorrectly originally

as-va: Yeah, that's what I assumed in the initial comment that something got lost in translation.

A payload in this context is what CMS sends us, which will not be the list of all services for a Facility, but individual service updates as individual services are updated.

taylorrassmannlibertyits: Behavior now updated

facilities/src/main/java/gov/va/api/lighthouse/facilities/CmsOverlayController.java
```diff
@@ -142,9 +143,62 @@ void updateCmsOverlayData(
...
+
+        List<DetailedService> currentDetailedServices =
+            cmsOverlayEntity.cmsServices() == null
+                ? Collections.emptyList()
+                : List.of(
+                    MAPPER_V0.readValue(cmsOverlayEntity.cmsServices(), DetailedService[].class));
+
+        List<DetailedService> toSaveDetailedServices = new ArrayList<>();
+
+        for (DetailedService currentDetailedService : currentDetailedServices) {
```
as-va, line 158: This looks like it would be simpler if instead of breaking up the active updates across these loops, instead have the first loop only get unchanged services (in existing list but not in new list) and the second loop add all active items in new list.

facilities/src/main/java/gov/va/api/lighthouse/facilities/CmsOverlayController.java
```diff
@@ -142,9 +143,62 @@ void updateCmsOverlayData(
...
+            if (currentDetailedService.name().equals(overlayService.name())
+                && overlayService.active()) {
+              toSaveDetailedServices.add(overlayService);
+              foundServiceToUpdate = true;
+              break;
+            }
+          }
+          // if we didn't find the service to update, this means it wasn't part of the payload
+          // and we need to persist the existing service data
+          if (!foundServiceToUpdate) {
```
as-va, line 172: This is incorrect. `foundServiceToUpdate` is only set to true if the incoming service is active. If the incoming service is _inactive_, it should not be persisted.

facilities/src/main/java/gov/va/api/lighthouse/facilities/CmsOverlayController.java
```diff
@@ -50,54 +107,160 @@ void initDirectFieldAccess(DataBinder dataBinder) {
...
+            facility.attributes().activeStatus(Facility.ActiveStatus.T);
+          } else {
+            facility.attributes().activeStatus(Facility.ActiveStatus.A);
+          }
+        }
 
-        // Since the covid 19 service name doesn't match our enum, we need to update and verify??
-        if (service.active()) {
+        if (overlay.detailedServices() != null) {
+          facility.attributes().detailedServices(activeServices.isEmpty() ? null : activeServices);
```
as-va, line 245: This will need the same updates as above to fix. Either break out the service update portion to a helper function both updates can call, or have the facility update pull from the already updated cms_overlay table instead of using the incoming overlay directly.

facilities/src/main/java/gov/va/api/lighthouse/facilities/CmsOverlayController.java
taylorrassmannlibertyits: I will work on updating this way

facilities/src/main/java/gov/va/api/lighthouse/facilities/CmsOverlayController.java
as-va: Just got feedback from CMS - they would like to keep the current behavior (`200` if overlay is applied to a facility, `201` if the facility is unknown).

taylorrassmannlibertyits: So this is not the best behavior, since we will be saving to the overlay table (with my current changes). A `200` response means request succeeded, but a `201` usually means success with a creation of a resource. In the old code no resource is created. We simply bail out. What do you suggest we do?

as-va: Ah, I mistyped - the code sends a `202` (HTTP: Accepted) when the facility ID is unknown.

taylorrassmannlibertyits: Ok, so do you want to still save the overlay if there is an unknown facility?

as-va: Yes, we store everything. If the stored overlay has a facility to apply to, return `200`; if the stored overlay does not (yet) have an associated facility, return `202`.

as-va: This shouldn't change anything already done in this PR except update the HTTP response code so it behaves like it did before the change.

taylorrassmannlibertyits: Ok I will update behavior

taylorrassmannlibertyits: There will be a ripple through some tests and some minor code changes.

facilities/src/main/java/gov/va/api/lighthouse/facilities/CmsOverlayController.java
as-va: Were you going to do this? This line has changed but it is still doing the lookup twice.

taylorrassmannlibertyits: The isPresent is required since it returns an optional entity. A warning is thrown if you just try and keep the entity.

as-va: This `isPresent` is needed, yes, but two hits to the DB is not. I have fixed in the latest push.

facilities/src/main/java/gov/va/api/lighthouse/facilities/CmsOverlayController.java
```diff
@@ -277,7 +227,15 @@ void updateFacilityData(
       }
       facilityEntity.overlayServices(detailedServices);
     }
-
     facilityRepository.save(facilityEntity);
   }
+
+  void updateServiceUrlPaths(String id, List<DetailedService> detailedServices) {
```
taylorrassmannlibertyits, line 233: This method was specifically named to replace covid 19 paths. I don't think this name should change

as-va (approved):
Approving as my initial change requests have been addressed, but as I was the last contributor to the branch, will wait for additional approvals before merging.

facilities/src/main/java/gov/va/api/lighthouse/facilities/CmsOverlayController.java
as-va: We only are actively supporting the vaccine service at this time, but any service we receive through this interface will need the path updated in the same way. This should be a generalized routine for all services, not one in particular.

taylorrassmannlibertyits: Based on this, then you need to rename the other information this method is handling and not just the name

facilities/src/main/java/gov/va/api/lighthouse/facilities/CmsOverlayController.java
```diff
@@ -34,9 +41,91 @@
...
+          overlay.detailedServices().stream()
+              .map(DetailedService::name)
+              .collect(Collectors.toList());
+    }
+    for (DetailedService currentDetailedService : currentDetailedServices) {
+      if (!overlayServiceNames.contains(currentDetailedService.name())) {
+        finalDetailedServices.add(currentDetailedService);
+      }
+    }
+    if (overlay.detailedServices() != null) {
```
esubocz-va, line 72: getActiveServicesFromOverlay method could be used here, and resulting list added to finalDetailedServices.

facilities/src/main/java/gov/va/api/lighthouse/facilities/CmsOverlayController.java
```diff
@@ -34,9 +41,91 @@
...
+      activeServices.sort(Comparator.comparing(DetailedService::name));
+    }
+    return activeServices;
+  }
+
+  @GetMapping(
+      value = {"/v0/facilities/{id}/cms-overlay", "/v1/facilities/{id}/cms-overlay"},
+      produces = "application/json")
+  @SneakyThrows
+  CmsOverlayResponse getOverlay(@PathVariable("id") String id) {
```
esubocz-va, line 104: this returns 200 OK when the overlay does not exist, 404 Not Found seems more appropriate.

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
```diff
@@ -361,6 +385,37 @@ private void moveToGraveyard(ReloadResponse response, FacilityEntity entity) {
...
+    // save entity
+    // done after all processing completes
+    boolean noErrors = true;
+    try {
+      log.warn("Attempting to save all facility overlay info to cms_overlay table.");
+      Streams.stream(facilityRepository.findAll())
+          .parallel()
+          .forEach(
+              f ->
+                  cmsOverlayRepository.save(
```
esubocz-va, line 402: when there has been no sms-overlay update, after this has been run, when I retrieve (for this test I used vha_523) the overlay, I get the response:
```
{
    "overlay": {}
}
```
there is no cms_overlay data for that Facility, but I would think that the ID should be populating from this, but it doesn't appear to be working properly.

facilities/src/main/java/gov/va/api/lighthouse/facilities/CmsOverlayController.java
as-va: I did not reuse that function because that would have caused two cycles through the path updating and sort functions which would add an unnecessary `O(n)` and `O(n log n)` complexity.

facilities/src/main/java/gov/va/api/lighthouse/facilities/CmsOverlayController.java
as-va: Agreed. Changed this to properly set HTTP status codes and return `200` when a valid overlay is returned and `404` if no overlay exists.

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
as-va: There is no ID in the response of a valid overlay _to_ populate, but this is still a good catch. `vha_523` is in the facility table, but `vha_000` is not. Neither ID has an overlay. Retrieving an overlay directly from the overlay table should have the same behavior if there is no overlay, but those two would have different results (the former returning an empty overlay and the latter returning a not found error). If there is no overlay, we should get back the `404` discussed above, not an empty overlay object. Updated this function to only migrate existing overlays from the facility table instead of creating empty overlays.

esubocz-va (approved):

PR #236, opened 2021-09-03
Api 8852 versioned data collection

esubocz-va:
Notes: 
- FacilityPair object removed
- DB table: facility column transitioned to facilityV1, V0 column added
- V0 transformer written & applied for populating DB.

Related story: https://vajira.max.gov/browse/API-8852


as-va:
Rejecting this PR as a better approach to V1 management has been implemented.

PR #242, opened 2021-10-04
API-10081 - Upgrade reload process to account for Covid URLs

mmccollum-va:
- Updates made to ensure that daily reload process, available at `/internal/management/reload` endpoint, recalculates Covid URL paths. 
- New internal endpoint `/internal/collector/overlays` introduced for purposes of providing a means to inform overlays to reprocess Covid URL paths.


giangnguyen2 (approved):

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalCollectorController.java
```diff
@@ -82,6 +104,44 @@
     }
   }
 
+  @PostMapping(value = "/overlays")
```
as-va, line 107: This breaks part of the pattern of this controller. This controller provides backend insight into the resources used during collection. It only load the various data sources and displays the source data (for debugging/maintenance); it does not change or modify any data or sources. Adding this method here changes that behavior and makes the data change by the mere process of viewing it.

It looks like the updating of the websites is also now done as part of the reload process. The URLs will now be updated 1) immediately upon reception of a new overlay and 2) during every daily reload cycle. That should be sufficient to keep the data maintained. If the purpose of this endpoint was to manually trigger a reload of the websites file, the reload endpoint can already do that. We already do an on-demand reload after a deployment that changes facility websites anyway so that would fit in to the existing process.

PR #243, opened 2021-10-04
API-9494 Correct Tests impacted by direct object comparison using FacilityEntity

giangnguyen2:
Update to ensure object comparison on FacilityEntity works correctly


mmccollum-va (dismissed):

facilities/src/test/java/gov/va/api/lighthouse/facilities/InternalFacilitiesControllerTest.java
```diff
@@ -332,7 +333,12 @@ void collect_createUpdate() {
     ReloadResponse response = _controller().reload().getBody();
     assertThat(response.facilitiesCreated()).isEqualTo(List.of("vha_f1"));
     assertThat(response.facilitiesUpdated()).isEqualTo(List.of("vha_f2"));
+    RecursiveComparisonConfiguration comparisonConfig =
```
as-va, line 336: This looks like the correct implementation to me - I think all the updates should look similar to this one.

facilities/src/test/java/gov/va/api/lighthouse/facilities/InternalFacilitiesControllerTest.java
```diff
@@ -710,7 +716,14 @@ void deleteFacilityOverlayById() {
     overlayRepository.save(_overlayEntity(_overlay(), "vha_f1"));
     response = _controller().deleteCmsOverlayById("vha_f1", "operating_status");
     assertThat(response.getStatusCode()).isEqualTo(HttpStatus.OK);
-    assertThat(facilityRepository.findAll()).isEqualTo(List.of(_facilityEntity(f, f1V1, null)));
+    RecursiveComparisonConfiguration facilityEntityCompConfig =
+        RecursiveComparisonConfiguration.builder()
+            .withIgnoredFields(
+                "version", "lastUpdated", "cmsOperatingStatus", "cmsServices", "overlayServices")
```
as-va, line 722: The extra fields added to the ignore list here are important fields and should not be ignored. As discussed earlier, the `version` and `lastUpdated` fields we do not control directly at the object level and cannot directly ensure that a before and after representation of the object would have the same values in these fields, so they are safe and logical to ignore. The overlay fields we *do* control and want to verify are the same before and after. If they are not, there is either a problem in the test setup or in the production code. These fields should come out of the ignore list and investigate why they aren't comparing the same in this test.

facilities/src/test/java/gov/va/api/lighthouse/facilities/InternalFacilitiesControllerTest.java
```diff
@@ -1088,8 +1103,16 @@ void upload() {
     facilityPairs.add(FacilityPair.builder().v0(f1).v1(f1V1).build());
     facilityPairs.add(FacilityPair.builder().v0(f2).v1(f2V1).build());
     _controller().upload(facilityPairs);
+    RecursiveComparisonConfiguration comparisonConfig =
+        RecursiveComparisonConfiguration.builder()
+            .withIgnoredFields("version", "lastUpdated")
+            .withIgnoreAllActualNullFields(true)
```
as-va, line 1109: As above, this should not be needed since we are already ignoring all the fields that are safe to ignore, so if this is necessary to make the test pass it may be masking some other problem. This should come out and figure out how to update the test setup to have the fields populated correctly or identify why in the production code this might be failing.

facilities/src/test/java/gov/va/api/lighthouse/facilities/InternalFacilitiesControllerTest.java
```diff
@@ -1088,8 +1103,16 @@ void upload() {
...
+            .build();
+    List<FacilityEntity> result = new ArrayList<>();
```
as-va, line 1111: This line and the next don't do anything for the test; looks like it may be leftover debug code. All debug code needs to come out before we merge.

facilities/src/test/java/gov/va/api/lighthouse/facilities/InternalFacilitiesControllerTest.java
giangnguyen2: The reason that `cmsOperatingStatus`, `cmsServices` and `cmsServices` are not the same for the `actual` and `expected` FacilityEntity is because in `deleteFacilityOverlayById` setup, the overlay was deleted using `InternalFacilitiesController.deleteCmsOverlayById`. This deletion will set `cmsOperatingStatus`, `cmsServices` and `cmsServices`  to null and cause mismatching. Because the test already explicitly does other assertions to compare `CmsOverlay` so I thought we wouldn't need to care about `CmsOverlay` when comparing `FacilityEntity`.

facilities/src/test/java/gov/va/api/lighthouse/facilities/InternalFacilitiesControllerTest.java
giangnguyen2: This has been changed

PR #248, opened 2021-10-29
Api 10768 detect multiple ids

ChrisBratti-VA:
Updated /reload resource to detect and report facilities comming in from VAST that are likely duplicates. This is determined based off of lat long coordinates, if facilities are within 0.2 miles of one another, they are considered duplicate faciclities and are flagged


facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/v0/ReloadResponse.java
```diff
@@ -47,11 +46,26 @@ public static ReloadResponse start() {
 
   @Builder
   @Value
-  @AllArgsConstructor(staticName = "of")
```
as-va, line 50: Instead of removing the `@AllArgsConstructor` and implementing all these yourself, you can leave the annotation in place and add only the one new form that you need. You can even chain it to use the existing forms by supplying a default.

```
public static Problem of(String facilityId, String description) {
    return Problem.of(facilityId, description, "");
}
```

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
```diff
@@ -90,6 +91,10 @@
 
   private final FacilityGraveyardRepository graveyardRepository;
 
+  private final Double tooClose = 0.02;
```
as-va, line 94: Nitpick mode:

This isn't a _bad_ name, but it could be better. Looking at this set of code changes, it is clear what this is for, but for another developer coming to this file later and not seeing only this diff but the whole file at once, this would need to be investigated to find out what it does. A name like 'duplicateDetectionRange' or `facilityOverlapThreshold` or similar make it more obvious at a glance what this does. This makes the code self-documenting.

I would also either add units to the variable name `...Miles` or add a comment with the declaration to give the units. Right now there's no way to know what the size or resolution of `0.02` means.

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
```diff
@@ -90,6 +91,10 @@
...
+
+  private final Set<FacilityEntity> facilities = new HashSet<>();
```
as-va, line 96: These are `FacilityEntity`s, so the variable name should also be `facilityEntities`, not `facilities` since _both_ are type names and this mismatch is misleading.

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
```diff
@@ -114,12 +119,15 @@ private static boolean isHoursNull(gov.va.api.lighthouse.facilities.api.v0.Facil
     return facility.attributes().hours() == null;
   }
 
+  private static Boolean isMobileCenter(FacilityEntity facility) {
+    return Optional.ofNullable(facility.mobile()).orElse(false)
+        && facility.id().stationNumber().endsWith("MVC");
```
as-va, line 124: I think we might be safer changing `endsWith` to `contains`. We've been _told_ that all Mobile Vet Centers end with `MVC`, but I don't know how rigorous that rule holds. If some Vet Center were to start operating two MVCs out of the same parent site, I don't know with certainty that wouldn't lead to `vc_###MVC1` and `vc_###MVC2` for instance.

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
```diff
@@ -698,4 +710,23 @@ private void updateFacility(ReloadResponse response, FacilityPair facilityPair)
     var response = ReloadResponse.start();
     return process(response, collectedFacilities);
   }
+
+  private List<String> within100FeetOfFacility(FacilityEntity newFacility) {
```
as-va, line 714: There are two problems with this function name: it uses different units (feet) than were established with the measurement variable above (miles) which is confusing **and** the name encodes specific logic in it that would make it harder to update in the future.

For point #1, I think we should stick with miles as the units and only reference feet parenthetically (e.g. `// Do something with 0.02 miles (~100 ft)`). I side with miles here because that is the unit _actually_ used in the haversine calculation that is being done and the feet number is a less precise approximation of that same value.

For point #2, what would happen if later on we decide that we are missing some potential duplicates and want to double the detection radius? Ideally, all we would have to do is update the class constant above, but here we would also have to update this function name, and make sure to update it everywhere it is used in the code. And if someone forgot to update the name, someone else would come along later and notice the constant and the class name don't agree and not know which is correct.

I think a more generalized but still specific name like `detectDuplicateFacilities` solves both those issues and keeps the code easily readable and self documented.

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
```diff
@@ -698,4 +710,23 @@ private void updateFacility(ReloadResponse response, FacilityPair facilityPair)
...
+    List<String> duplicateIds = new ArrayList<>();
+    getAllFacilities()
```
as-va, line 716: This logic works, but you can also use Java streams to simplify this quite a bit.

```
facilities.stream()
    .filter(f -> f.id().type() == newFacility.id().type())
    .filter(f -> !f.id().stationNumber().equals(newFacility.id().stationNumber()))
    .filter(f -> !isMobileCenter(f) && !isMobileCenter(newFacility))
    .filter(f -> FacilityUtils.haversine(newFacility, f.longitude(), f.latitude()) <= tooClose)
    .map(f -> f.id().toIdString())
    .forEachOrdered(duplicateIds::add);
```

as-va (changes requested):

as-va:
There are no unit test changes associated with this PR. I assume that with this new functionality additional testing will be necessary. This may be updates to existing tests or new tests created.

as-va (approved):

PR #250, opened 2021-11-02
API-10475-enable-facilities-data-to-be-api-version-agnostic-and-update-controllers-accordingly

mmccollum-va:
`DatamartFacility` has been introduced to enable the persistence of facility data to be API version agnostic. The various controllers have been updated to use a versioned facility transformer utility class to transform agnostic facility data to a versioned facility object for external representation and back. This enables the JSON facility data to be persisted independent of how the facility data is represented externally.


facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/v1/CmsOverlay.java
```diff
@@ -0,0 +1,24 @@
...
+import lombok.Builder;
+import lombok.Data;
+
+@Data
+@Builder
+@JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY)
+@Schema(description = "Data provided by CMS to Facilities to be applied on top of known data.")
+public class CmsOverlay {
+  @Valid
+  @JsonProperty("operating_status")
```
bschofield-va, line 18: You could replace these with `@JsonNaming(PropertyNamingStrategy.SnakeCaseStrategy)` on the class. Same net result, so you do you.

bschofield-va: oh ... according to the most recently published styleguide ... these should follow camelCase.
https://legendary-system-8b212937.pages.github.io/unified/naming-conventions/

facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/v1/CmsOverlay.java
```diff
@@ -0,0 +1,24 @@
...
+  Facility.OperatingStatus operatingStatus;
+
+  @Valid
```
bschofield-va, line 21: `@Valid List<Whatever> whatever` is legal and supported for backwards compatibility,but Bean Validation revised how this should be done with the 2.0 specification. This would be better as `List<@Valid Whatever> whatever`

https://beanvalidation.org/2.0/spec/#constraintdeclarationvalidationprocess-containerelementconstraints

facilities/src/main/java/gov/va/api/lighthouse/facilities/CmsOverlayControllerV1.java
```diff
@@ -0,0 +1,238 @@
+package gov.va.api.lighthouse.facilities;
```
bschofield-va, line 1: This is pretty much a copy and paste of `facilities/src/main/java/gov/va/api/lighthouse/facilities/CmsOverlayControllerV0.java` 
That's a lot of duplication. I'd advice in extracting that logic into a single, common, reusable place.

facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/v1/CmsOverlay.java
```diff
@@ -0,0 +1,24 @@
...
+import lombok.Data;
+
+@Data
+@Builder
+@JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY)
+@Schema(description = "Data provided by CMS to Facilities to be applied on top of known data.")
+public class CmsOverlay {
+  @Valid
+  @JsonProperty("operating_status")
+  Facility.OperatingStatus operatingStatus;
```
giangnguyen2, line 19: There another currently outstanding pull request [here](https://github.com/department-of-veterans-affairs/lighthouse-facilities/pull/236) that factor `OperatingStatus` into its own class. Not sure which pull request will take precedence and it might cause some code conflict.

facilities/src/main/java/gov/va/api/lighthouse/facilities/CmsOverlayControllerV1.java
mmccollum-va: Reusable logic extracted to newly created abstract parent class for CMS overlay controllers.

facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/v1/CmsOverlay.java
mmccollum-va: Updated constraints applied to elements of generic containers according to Bean Validation 2.0 spec

facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/v1/CmsOverlay.java
mmccollum-va: Updated to use camelCase naming convention for JSON properties according to latest published styleguide spec

facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/v1/CmsOverlay.java
as-va: The referenced branch may be usable for extracting portions of the changes, but the overall direction of that change was towards _furthering_ version specific semantics in the API/DB whereas this PR is driving _away_ from that approach. Once this PR is merged, the other will be closed out without merging.

facilities/src/main/java/gov/va/api/lighthouse/facilities/BaseCmsOverlayController.java
```diff
@@ -0,0 +1,64 @@
...
+      String id,
+      List<DetailedService> detailedServices,
+      ObjectMapper mapper) {
+    List<DetailedService> currentDetailedServices =
+        cmsOverlayEntity.cmsServices() == null
+            ? Collections.emptyList()
+            : List.of(mapper.readValue(cmsOverlayEntity.cmsServices(), DetailedService[].class));
+    List<DetailedService> finalDetailedServices = new ArrayList<>();
+    List<String> overlayServiceNames = new ArrayList<>();
+    if (detailedServices != null) {
```
bschofield-va, line 27: imho, nullable collections are bad as it leads to
- easy null pointer mistakes, e.g., `if ( someList.isEmpty() )`
- and ugly code, e.g., `if ( someList != null )`

I advise minimizing the potential for nullable lists, sets, maps, etc. and go so far as to say that in general never allow nullable collections.

For example, you could 
- add a `@NonNull List<DetailedService>` would remove two conditionals from this method and force the callers to provide an empty (`List.of()`) 
- or immediately replace null list with an safe, empty list at the beginning of this method.

facilities/src/main/java/gov/va/api/lighthouse/facilities/CmsOverlayControllerV1.java
```diff
@@ -0,0 +1,188 @@
...
+    FacilityEntity.Pk pk = FacilityEntity.Pk.fromIdString(id);
+    Optional<CmsOverlayEntity> existingOverlayEntity = cmsOverlayRepository.findById(pk);
+    if (existingOverlayEntity.isPresent()) {
+      CmsOverlayEntity cmsOverlayEntity = existingOverlayEntity.get();
+      CmsOverlayResponse response =
+          CmsOverlayResponse.builder()
+              .overlay(
+                  CmsOverlay.builder()
+                      .operatingStatus(
+                          cmsOverlayEntity.cmsOperatingStatus() == null
```
bschofield-va, line 58: why not put this "null or read value" logic behind a clean method that allows you to express business logic intention more clearly and keep the details of the ternary in a lower level of abstraction?, e.g. ?

```
.operatingStatus(nullableValue(cmsOverlayEntity.cmsOperatingStatus(), Facility.OperatingStatus.class))
```

facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/CmsOverlayCollectorV1.java
```diff
@@ -0,0 +1,95 @@
...
+    Streams.stream(cmsOverlayRepository.findAll())
+        .parallel()
+        .forEach(
+            cmsOverlayEntity -> {
+              try {
+                returnMap.put(
+                    cmsOverlayEntity.id().toIdString(),
+                    CmsOverlay.builder()
+                        .operatingStatus(
+                            cmsOverlayEntity.cmsOperatingStatus() != null
```
bschofield-va, line 53: here's more of the `if not null , read value` logic that could be encapsulated nicely

facilities/src/main/java/gov/va/api/lighthouse/facilities/DatamartFacility.java
```diff
@@ -0,0 +1,372 @@
...
+
+@Data
+@Builder
+@JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY)
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+@AllArgsConstructor(access = AccessLevel.PRIVATE)
+public class DatamartFacility {
+  @NotNull String id;
+
+  @NotNull DatamartFacility.Type type;
```
as-va, line 28: This file is full of `DatamartFacility.[field]` prefixes on fields that should not be necessary since they are all local to this file and there is no ambiguity in type resolution. This may have been added by the compiler as a result of a copy/paste from the source `/v0/Facility.java` file?

facilities/src/main/java/gov/va/api/lighthouse/facilities/DatamartFacility.java
```diff
@@ -0,0 +1,372 @@
...
+    String thursday;
+
+    String friday;
+
+    String saturday;
+
+    String sunday;
+
+    public static final class HoursBuilder {
+      @JsonProperty("Friday")
```
as-va, line 236: The days ordering here is an artifact of the old Ruby version of the API that we no longer need to persist. Let's make the ordering here follow natural ordering like the fields above do.

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
```diff
@@ -82,7 +82,7 @@
 
   //  private static final ObjectMapper MAPPER_V1 = FacilitiesJacksonConfigV1.createMapper();
 
-  private final FacilitiesCollector collector;
+  private final FacilitiesCollectorV0 collector;
```
as-va, line 85: This seems as good a place to leave this comment as any - this is a general architecture observation that doesn't really change _this_ line, but the is the top of the chain of logic.

We have split our data classes into V0 and V1 containers to hold different data and our controllers into V0 and V1 variants to support different operations, but our backend data objects are singular entities that just get transformed to different views. We have a single DatamartFacility (list of) object which is populated by a single reload endpoint (`/internal/management/reload`) that collects the single set of data that comprise the object, so there really is no need for V0/V1 bifurcation at this level. Put another way, the collection process has to result in the same data being populated in the database, so there is no need for multiple collection pathways. Only the V0 collector is being used anyway, so this V0 collector can stay a base data type collector. This should only require changing the final stage of the subcollectors from a `toFacility` mapping to a `toDatamartFacility`. The `collectV1` functions in the collectors can be removed and the call chain updated to handle `List<DatamartFacility>` rather than `List<FacilityPair>` and the later container type removed.

facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilityTransformerV0.java
```diff
@@ -0,0 +1,513 @@
...
+                        transformFacilityOperatingStatus(f.attributes().operatingStatus()))
+                    .detailedServices(f.attributes().detailedServices())
+                    .operationalHoursSpecialInstructions(
+                        f.attributes().operationalHoursSpecialInstructions())
+                    .build()
+                : null)
+        .build();
+  }
+
+  /** Transform DatamertFacility active status to version 0 facility active status. */
```
as-va, line 88: Datam`e`rtFacility typos throughout both FacilityTransformer files.

facilities/src/main/java/gov/va/api/lighthouse/facilities/BaseCmsOverlayController.java
mmccollum-va: This has been addressed: null list immediately replaced with a safe, empty list at beginning of method.

facilities/src/main/java/gov/va/api/lighthouse/facilities/CmsOverlayControllerV1.java
mmccollum-va: "null or read value" logic encapsulated in its own method

facilities/src/main/java/gov/va/api/lighthouse/facilities/CmsOverlayControllerV1.java
```diff
@@ -0,0 +1,188 @@
...
+  private final CmsOverlayRepository cmsOverlayRepository;
+
+  @GetMapping(
+      value = {"/v1/facilities/{id}/cms-overlay"},
+      produces = "application/json")
+  @SneakyThrows
+  ResponseEntity<CmsOverlayResponse> getOverlay(@PathVariable("id") String id) {
+    FacilityEntity.Pk pk = FacilityEntity.Pk.fromIdString(id);
+    Optional<CmsOverlayEntity> existingOverlayEntity = cmsOverlayRepository.findById(pk);
+    if (existingOverlayEntity.isPresent()) {
```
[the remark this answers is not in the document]
mmccollum-va: Level of indent removed

facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/CmsOverlayCollectorV1.java
mmccollum-va: "null or read value" logic encapsulated in its own method

facilities/src/main/java/gov/va/api/lighthouse/facilities/DatamartFacility.java
mmccollum-va: This has been addressed

facilities/src/main/java/gov/va/api/lighthouse/facilities/DatamartFacility.java
mmccollum-va: Code formatting performed by maven build determines display order of methods.

facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilityTransformerV0.java
mmccollum-va: Typos corrected.

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
mmccollum-va: API implementation use of `FacilityPair` minimized by refactoring subcollectors for benefits, cemeteries, healths, state cemeteries, and vet centers to utilize API version agnostic `DatamartFacility`. Bifurcated V0/V1 transformers for these subcollectors are now somewhat redundant and will be addressed in future commit. `FacilitiesCollector` maintains reference to `CmsOverlayCollector`, which itself maintains versioned `Facility` reference. Both `FacilityCollector` and `CmsOverlayCollector` remain bifurcated into V0/V1 versions due to this referenced to versioned `Facility` object.

facilities/src/main/java/gov/va/api/lighthouse/facilities/BaseCmsOverlayController.java
```diff
@@ -0,0 +1,50 @@
...
+    finalDetailedServices.sort(Comparator.comparing(DetailedService::name));
+    return finalDetailedServices;
+  }
+
+  protected List<DetailedService> getActiveServicesFromOverlay(
+      String id, List<DetailedService> detailedServices) {
+    final List<DetailedService> activeServices = new ArrayList<>();
+    if (detailedServices != null) {
+      detailedServices.parallelStream().filter(d -> d.active()).forEach(d -> activeServices.add(d));
+    }
```
giangnguyen2, line 45: Adding element to thread-unsafe ArrayList in a parallel stream can cause unexpected behavior. I think it might be better to use `Collectors.toList()` instead of `forEach(d -> activeServices.add(d));`

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
mmccollum-va: Redundant transformers for benefits, cemeteries, healths, state cemeteries, and vet centers have been marked as deprecated, commented out, and their usage removed.

facilities/src/main/java/gov/va/api/lighthouse/facilities/BaseCmsOverlayController.java
mmccollum-va: Good catch. This has been addressed.

facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilityTransformerV1.java
```diff
@@ -51,8 +50,27 @@ static Facility toFacility(@NonNull DatamartFacility df) {
         .build();
   }
 
+  /** Transform DatamertFacility operating status to version 1 facility operating status. */
```
as-va, line 53: This file still contains Datam`e`rtFacility in comments.

mmccollum-va: Typos addressed

facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/BenefitsTransformerV1.java
```diff
@@ -1,142 +1,189 @@
...
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Locale;
 import lombok.Builder;
-import lombok.NonNull;
-import org.apache.commons.lang3.StringUtils;
 
+@Deprecated
 @Builder
 final class BenefitsTransformerV1 {
```
as-va, line 65: This file is not used anywhere (and has no contents) so it can be removed from the project.

facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/CemeteriesTransformerV1.java
```diff
@@ -1,90 +1,99 @@
...
+// import gov.va.api.lighthouse.facilities.DatamartFacility.Phone;
+// import java.util.Locale;
+// import lombok.NonNull;
 
-import gov.va.api.lighthouse.facilities.api.v1.Facility;
-import java.util.Locale;
 import lombok.Builder;
-import lombok.NonNull;
 
+@Deprecated
```
as-va, line 21: This can also be removed as per above.

facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/HealthTransformerV1.java
```diff
@@ -1,400 +1,432 @@
...
-import java.util.List;
-import java.util.Locale;
-import java.util.Map;
-import java.util.Objects;
-import java.util.TreeMap;
 import lombok.Builder;
-import lombok.NonNull;
-import org.apache.commons.lang3.BooleanUtils;
 
+@Deprecated
```
as-va, line 67: This can also be removed as per above.

facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/StateCemeteryTransformerV1.java
```diff
@@ -1,192 +1,194 @@
...
-import java.util.Locale;
-import java.util.Map;
-import java.util.Objects;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
-import java.util.stream.Stream;
 import lombok.Builder;
-import lombok.NonNull;
 
+@Deprecated
```
as-va, line 30: This can also be removed as per above.

facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/VetCenterTransformerV1.java
```diff
@@ -1,148 +1,156 @@
...
+// import java.util.Map;
+// import lombok.NonNull;
 
-import gov.va.api.lighthouse.facilities.api.v1.Facility;
-import java.util.Locale;
-import java.util.Map;
 import lombok.Builder;
-import lombok.NonNull;
 
+@Deprecated
```
as-va, line 25: This can also be removed as per above.

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
```diff
@@ -127,31 +130,25 @@ private static Boolean isMobileCenter(FacilityEntity facility) {
...
-    record.visn(facilityV0.attributes().visn());
-    record.mobile(facilityV0.attributes().mobile());
-    // gov.va.api.lighthouse.facilities.api.v1.Facility facilityV1 = facilityPair.v1();
-    // record.facilityV1(MAPPER_V1.writeValueAsString(facilityV1));
+    record.latitude(datamartFacility.attributes().latitude().doubleValue());
+    record.longitude(datamartFacility.attributes().longitude().doubleValue());
+    record.state(stateOf(datamartFacility));
+    record.zip(zipOf(datamartFacility));
+    record.servicesFromServiceTypes(serviceTypesOf(datamartFacility));
+    record.facility(MAPPER_V0.writeValueAsString(datamartFacility));
```
as-va, line 140: Why are we still using a `V0` mapper object for a `DatamartFacility` type?

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
```diff
@@ -704,19 +734,23 @@ private void updateFacility(ReloadResponse response, FacilityPair facilityPair)
...
-    log.warn("Creating new facility {}", facility.id());
-    updateAndSave(response, FacilityEntity.builder().id(pk).build(), facilityPair);
+    response.facilitiesCreated().add(datamartFacility.id());
+    log.warn("Creating new facility {}", datamartFacility.id());
+    updateAndSave(response, FacilityEntity.builder().id(pk).build(), datamartFacility);
   }
 
   @PostMapping(value = "/reload")
   @Loggable(arguments = false)
   ResponseEntity<ReloadResponse> upload(@RequestBody List<FacilityPair> collectedFacilities) {
```
as-va, line 748: This should receive a `DatamartFacility`.

facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/BenefitsTransformerV1.java
mmccollum-va: `BenefitsTransformerV1` and its corresponding test have been removed.

facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/CemeteriesTransformerV1.java
mmccollum-va: `CemeteriesTransformerV1` and its corresponding test have been removed.

facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/HealthTransformerV1.java
mmccollum-va: `HealthTransformerV1` and its corresponding test have been removed.

facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/StateCemeteryTransformerV1.java
mmccollum-va: `StateCemeteryTransformerV1` and its corresponding test have been removed.

facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/VetCenterTransformerV1.java
mmccollum-va: `VetCenterTransformerV1` and its corresponding test have been removed.

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalCollectorController.java
```diff
@@ -52,7 +53,14 @@
 
   @GetMapping(value = "/facilities")
   List<FacilityPair> collectFacilities() {
```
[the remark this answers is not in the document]
mmccollum-va: Use of `FacilityPair` has been removed and internal collector returns `List<DatamartFacility>`

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
mmccollum-va: `upload` now receives a `DatamartFacility`

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalCollectorController.java
```diff
@@ -22,11 +23,11 @@
 public class InternalCollectorController {
   private final JdbcTemplate jdbc;
 
-  private final FacilitiesCollector collector;
+  private final FacilitiesCollectorV0 collector;
```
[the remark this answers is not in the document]
mmccollum-va: For now, `FacilitiesCollectorV0` must remain versioned as it still maintains a versioned reference to CMS overlay via `CmsOverlayCollectorV0`. Future work to create a version agnostic CMS overlay should help remediate this.

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
mmccollum-va: Custom object mapper created for `DatamartFacility` objects.

as-va (approved):

PR #251, opened 2021-11-08
API-11212-Updating-URLS

JamesDanielBell-VA:
https://vajira.max.gov/browse/API-11212

- Updated Facility URLs


facilities/src/main/resources/websites.csv
```diff
@@ -272,15 +272,19 @@ vc_0608V,https://www.va.gov/temecula-vet-center/
...
 vc_0619V,https://www.va.gov/san-luis-obispo-vet-center/
 vc_0620V,https://www.va.gov/san-francisco-vet-center/
 vc_0623V,https://www.va.gov/east-los-angeles-vet-center/
-vc_0624V,https://www.va.gov/san-marcos-vet-center/
+vc_0624V,https://www.va.gov/north-orange-county-vet-center/
+vc_0626V,https://www.va.gov/eugene-vet-center/
 vc_0628V,https://www.va.gov/fresno-vet-center/
 vc_0634V,https://www.va.gov/maui-vet-center/
 vc_0637V,https://www.va.gov/san-bernardino-vet-center/
+vc_0638V,https://www.va.gov/sacramento-vet-center
```
as-va, line 287: This matches what was in the requested changes, but this line is missing a trailing slash which nearly all other VCs have. To be consistent, let's add a slash to the end of this line as well as to the other three VC sites that don't have one
* vc_0518V
* vc_0609V
* vc_0833MVC

as-va (approved):

PR #253, opened 2021-11-12
API-11028 F: Modify Agent J Updater to Update drive_time_band Table all at once

giangnguyen2:
Added new endpoint to return distinct band versions that will be utilized by agent-j version checker to determine if there was a partial update occurred


as-va (approved):

as-va (changes requested):
This PR adds new code but there is no additional unit testing provided. Please add unit test coverage of new features.

as-va (approved):

PR #259, opened 2021-12-03
API-11635 F: Vet Center URL Updates

giangnguyen2:


facilities/src/main/resources/websites.csv
```diff
@@ -278,6 +293,8 @@ vc_0612V,https://www.va.gov/oakland-vet-center/
 vc_0613V,https://www.va.gov/high-desert-vet-center/
 vc_0614V,https://www.va.gov/chula-vista-vet-center/
 vc_0615V,https://www.va.gov/san-jose-vet-center/
+vc_0616v,https://www.va.gov/american-samoa-vet-center
```
as-va, line 296: Curious that this one has a lower case `v`. The actual ID for this site in the API has a capital so let's update this to be consistent.

as-va: Also, this line, the following, and one more further down are missing a `/` at the end of the path. This makes no difference as far as reachability of the link, but we should add it to these three so they stay consistent with the formatting of the rest of the links in the API.

giangnguyen2: Updated lower case v  to V and added  `/` to the end of URLs

as-va: I see the slashes now, but the `v` is still lowercase. Is there a commit that didn't get pushed?

giangnguyen2: My bad. I  updated the supplied url files and re-ran the program not realizing that the TreeMap put method use case-insensitive comparison so the `v` didn't get updated. It should be good now

as-va (approved):

PR #264, opened 2021-12-08
API-8856 F: V1: Deprecate graveyard

giangnguyen2:
Remove graveyard from facilities API [API-8856](https://vajira.max.gov/browse/API-8856)


facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
```diff
@@ -484,7 +391,6 @@ private void processMissingFacility(ReloadResponse response, FacilityEntity.Pk i
       saveAsMissing(response, entity);
       return;
     }
-    moveToGraveyard(response, entity);
```
as-va, line 487: In addition to the graveyard stuff, this function also deleted the entity from the `facilityRepository` so that still needs to be done here.

as-va (changes requested):

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
giangnguyen2: Good catch. I have added `facilityRepository.delete(entity)` to delete facility

as-va (approved):

PR #265, opened 2021-12-08
API-10477 F: V1: Create service collection endpoint (CMS work)

mmccollum-va:
Introduction of new Facilities API endpoint to return collection of services available for a `Facility`.


giangnguyen2 (dismissed):

giangnguyen2 (approved):

facilities/src/main/java/gov/va/api/lighthouse/facilities/CmsOverlayControllerV1.java
```diff
@@ -42,13 +42,26 @@
 
   private final CmsOverlayRepository cmsOverlayRepository;
 
+  @GetMapping(
+      value = {"/v1/facilities/{id}/services"},
+      produces = "application/json")
+  @SneakyThrows
+  ResponseEntity<List<DetailedService>> getDetailedServices(@PathVariable("id") String id) {
```
as-va, line 49: This is a good start but we need to "API-ify" it a bit. Instead of just returning a list of json objects, this should return a `ServiceResponse` type object which contains a list of service objects in a data node. This will allow us to modify the response object with additional nodes if needed without breaking or altering the `DetailedService` type. As this endpoint returns a collection of objects, it also needs to support pagination. See `FacilitiesReponse` and `FacilityReadResponse` for the Facilities equivalents.

mmccollum-va: New CMS endpoint updated to return `DetailedServicesResponse` object containing list of `DetailedService`s, pager links, and associated metadata.

as-va (approved):

PR #266, opened 2021-12-09
API-10478 - F: V1: Create service item endpoint (CMS Work)

mmccollum-va:
Introduction of new endpoint to return a single service associated with a `Facility`, as part of the CMS Health Service integration.

Note: Changes from `API-10477` have been merged into this branch.


facilities/src/main/java/gov/va/api/lighthouse/facilities/CmsOverlayControllerV1.java
```diff
@@ -42,13 +43,36 @@
 
   private final CmsOverlayRepository cmsOverlayRepository;
 
+  @GetMapping(
+      value = {"/v1/facilities/{facility_id}/services/{service_id}"},
+      produces = "application/json")
+  @SneakyThrows
+  ResponseEntity<DetailedService> getDetailedService(
```
as-va, line 50: This is a good start but we need to "API-ify" it a bit. Instead of just returning a json object, this should return a ServiceResponse type object which contains a service object in a data node. This will allow us to modify the response object with additional nodes if needed without breaking or altering the DetailedService type. As this endpoint only returns a single object (not a collection), it does not need to support pagination. See FacilitiesReponse and FacilityReadResponse for the Facilities equivalents.

mmccollum-va: New CMS endpoint updated to return `DetailedServiceResponse` object containing user requested `DetailedService`

facilities/src/main/java/gov/va/api/lighthouse/facilities/BaseCmsOverlayController.java
```diff
@@ -56,13 +57,22 @@
   protected abstract Optional<CmsOverlayEntity> getExistingOverlayEntity(FacilityEntity.Pk pk);
 
   @SneakyThrows
-  protected List<DetailedService> getOverlayDetailedServices(String id) {
-    FacilityEntity.Pk pk = FacilityEntity.Pk.fromIdString(id);
+  protected DetailedService getOverlayDetailedService(
+      @NonNull String facilityId, @NonNull String serviceId) {
+    List<DetailedService> detailedServices =
+        getOverlayDetailedServices(facilityId).parallelStream()
+            .filter(ds -> ds.name().equals(serviceId))
```
as-va, line 64: Let's have this be `equalsIgnoreCase`. I don't like the thought of false negatives for searches for "Covid-19 vaccines" when the service name is "COVID-19 vaccines".

as-va (approved):

PR #269, opened 2021-12-17
API-11979 - F: V1: Baseline cleanup work

mmccollum-va:
Code cleanup of Facilities API V1 baseline.


facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilitiesJacksonConfigV1.java
```diff
@@ -11,17 +12,17 @@ public static ObjectMapper createMapper() {
...
-  //  }
-  //
-  //  /** Mask away checked exception so this Jackson can be used in streams. */
-  //  @SneakyThrows
-  //  static String quietlyWriteValueAsString(ObjectMapper mapper, Object obj) {
-  //    return mapper.writeValueAsString(obj);
-  //  }
+  /** Mask away checked exception so this Jackson can be used in streams. */
+  @SneakyThrows
+  static <T> T quietlyMap(ObjectMapper mapper, String json, Class<T> type) {
```
as-va, line 17: There is an additional `quietlyMap` function that was added to the V0 after this branched off from it that we may need to pull over.

facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilityOverlayV1.java
```diff
@@ -1,97 +1,110 @@
...
+import lombok.extern.slf4j.Slf4j;
+
+@Builder
+@Value
+@Slf4j
+public class FacilityOverlayV1 implements Function<HasFacilityPayload, Facility> {
+
+  @NonNull ObjectMapper mapper;
+
+  private static void applyCmsOverlayOperatingStatus(
```
as-va, line 26: This function and `applyDetailedServices` below no longer exist in V0 and the usages in `apply` (also in this file) have been removed. Those changes likely need ported here as well to still play nice with the recent data layer changes.

as-va (approved):

PR #271, opened 2021-12-28
API-7323 F: V1 Normalize days/hours mapping in facilities

giangnguyen2:
Standardize the day/hours mapping in facilities:
Monday - Sunday order 
Capitalize First letter of day.
[API-7323](https://vajira.max.gov/browse/API-7323)


facilities/src/main/java/gov/va/api/lighthouse/facilities/JacksonSerializersV1.java
```diff
@@ -273,26 +273,6 @@ private static boolean empty(Facility.Hours value) {
...
-      writeNonNull(jgen, "friday", value.friday());
-      writeNonNull(jgen, "monday", value.monday());
-      writeNonNull(jgen, "sunday", value.sunday());
-      writeNonNull(jgen, "tuesday", value.tuesday());
-      writeNonNull(jgen, "saturday", value.saturday());
-      writeNonNull(jgen, "thursday", value.thursday());
-      writeNonNull(jgen, "wednesday", value.wednesday());
-    }
-
     private static void writeUpper(Facility.Hours value, JsonGenerator jgen) {
```
as-va, line 276: Instead of modifying the old `HoursSerializer`, this whole class can be deleted. It was needed in V0 since the serialization had to be different depending on which generator type was using it, but in V1 all objects serialize hours in the same way. You have already annotated the V1/Facility.Hours class to specify the JsonPropertyOrder of the fields so there is no need to overwrite the default serialization behavior here.

giangnguyen2: Cool. I will remove it. I was debating it too :)

as-va (approved):

PR #272, opened 2022-01-03
API-11367 F: V0: Add radius filtering to Lat/Lon searches

mmccollum-va:
Addition of `radius` filtering to both V0 and V1 lat/lon searches.


facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/FacilitiesSearchApi.java
```diff
@@ -24,12 +24,14 @@
   @Operation(
       summary = "Query facilities by location or IDs, with optional filters",
       description =
-          "Query facilities by bounding box, latitude and longitude, state, visn, or zip code. "
-              + "Bounding box is specified as four `bbox[]` parameters, long1, lat1, long2, lat2. "
-              + "(Relative order is unimportant.)"
+          "Query facilities by bounding box, latitude and longitude, radius, state, visn, or "
```
as-va, line 27: I think we should leave `radius` out of this line. This sentence enumerates the search modes that are offered - radius is an optional parameter more like services and mobile which are called out below. This implies there is a distinct radius search mode orthogonal to lat/long.

The rest of the documentation updates here look good.

mmccollum-va: Beginning statement clarified to ensure that there is no implication of a distinct radius search mode orthogonal to lat/lon.

as-va (approved):

PR #273, opened 2022-01-04
Api 5566 remove pipes clean

JamesDanielBell-VA:
https://vajira.max.gov/browse/API-5566

- Updated operationalHoursSpecialInstructions from string to list.
- Updated test classes to account for string to list change.
- Added methods to transform operationalHoursSpecialInstructions from list to string and string to list.


facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/v1/Facility.java
```diff
@@ -205,10 +205,15 @@
     Hours hours;
 
     @Schema(
-        example = "Normal business hours are Monday through Friday, 8:00 a.m. to 4:30 p.m.",
+        example =
+            "\"operational_hours_special_instructions\" : ["
+                + "\"Vet center 1 is available.\","
```
as-va, line 210: Since this is the documentation that is published for consumers to see what a response object looks like, let's make this examples of real data.

as-va (approved):

PR #277, opened 2022-01-11
Api 11854 enhance duplicate facility logic

JamesDanielBell-VA:
https://vajira.max.gov/browse/API-11854

-Updated InternalFacilitiesController : isMobile method to run on health facilities as well as vet centers
-Updated InternalFacilitiesController to sort the Duplicate Id's that are collected and output to CSV
-Updated agent-j : build.sh to count unique duplicate id rows instead of all possible permutations


giangnguyen2 (approved):

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
```diff
@@ -119,7 +120,8 @@ private static boolean isHoursNull(DatamartFacility datamartFacility) {
 
   private static Boolean isMobileCenter(FacilityEntity facility) {
     return Optional.ofNullable(facility.mobile()).orElse(false)
-        && facility.id().stationNumber().contains("MVC");
+        && (facility.id().stationNumber().contains("MVC")
```
as-va, line 123: Let's just remove these two lines and just look at the mobile flag. This logic is getting obtuse and `Vet Centers` and `Health` facilities are the *only* types that can be mobile anyway so these checks are superfluous. If we ever end up with a mobile cemetery we'll have bigger things to worry about!

as-va (approved):

PR #283, opened 2022-01-28
API-12740 F: V1: Register service names in service enumeration

mmccollum-va:
List of service Ids added to enumerated list to enable population on VA Developer Portal.

- [ ] Resolve prior to merging: 
`SpecialtyCare` health service was temporarily added back to FAPI V1 `Facility`. `HealthTransformer` converts health services from ATC to CMS. ATC views non-`PrimaryCare` services as `SpecialtyCare`. CMS contains a variety of specialized health services but does not contain a vanilla `SpecialtyCare` bucket covering them all. Awaiting additional feedback on mapping between ATC `SpecialtyCare` and CMS specialized services.


facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/HealthTransformer.java
```diff
@@ -89,12 +88,11 @@
     map.put("DERMATOLOGY", Dermatology);
     map.put("GASTROENTEROLOGY", Gastroenterology);
     map.put("GYNECOLOGY", Gynecology);
-    map.put("MENTAL HEALTH", MentalHealthCare);
+    map.put("MENTAL HEALTH", MentalHealth);
     map.put("OPHTHALMOLOGY", Ophthalmology);
     map.put("OPTOMETRY", Optometry);
     map.put("ORTHOPEDICS", Orthopedics);
     map.put("PRIMARY CARE", PrimaryCare);
-    map.put("SPECIALTY CARE", SpecialtyCare);
```
as-va, line 97: This map is used for translating ATC's service names to our service identifiers. Won't this change prevent us from pulling the service and associated wait_time?

mmccollum-va: Correct, this was initially removed, but has been restored.

as-va (approved):

PR #288, opened 2022-02-04
API-13193 Follow-on to registering service names in FAPI V1 service enumeration

mmccollum-va:
Changes made to ensure Jackson can roundtrip all `HealthService` versions regardless of capitalization.


facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/v0/Facility.java
```diff
@@ -81,7 +84,15 @@
     SpecialtyCare,
     UrgentCare,
     Urology,
-    WomensHealth
+    WomensHealth;
+
+    /** Ensure that Jackson can create HealthService enum regardless of capitalization. */
+    @JsonCreator
```
ChrisBratti-VA, line 90: How is the fromString() method being used? I'm assuming its for deserialization but I didn't see it called anywhere. Does the `@JsonCreator` annotation provide some sort of built-in functionality from Jackson?

mmccollum-va: Correct, usage of `fromString` method when deserializing enum JSON string is controlled by `@JsonCreator` annotation

ChrisBratti-VA (approved):

facilities/src/test/java/gov/va/api/lighthouse/facilities/BaseFacilityTransformerTest.java
```diff
@@ -0,0 +1,41 @@
+package gov.va.api.lighthouse.facilities;
+
+import static gov.va.api.health.autoconfig.configuration.JacksonConfig.createMapper;
+
+import gov.va.api.lighthouse.facilities.api.v0.Facility;
```
as-va, line 5: For common classes that need to reference both versioned Facility types, I prefer referencing both with long form notation and only importing a Facility type name when that is the only Facility type reference in the file. Importing only one causes the code to have `Facility` for V0 and `g.v.a.l.f.a.v1.Facility` for V1 which gives the false impression that this is a "V0" file with a V1 Facility bolted on instead of a class that treats both objects equally.

as-va (approved):

PR #289, opened 2022-02-07
API-12980 F: CDW request to decrease connection sessions to SQL52

giangnguyen2:
Set minimum number of connection, maximum number of connection, and idle timeout for facility collector
[API-12980](https://vajira.max.gov/browse/API-12980)


facilities/src/main/resources/application.properties
```diff
@@ -25,10 +25,14 @@ spring.datasource.password=unset
...
 spring.datasource.username=unset
 
+#Facilities collector connection pooling properties
 spring.fc-datasource.driver-class-name=com.microsoft.sqlserver.jdbc.SQLServerDriver
 spring.fc-datasource.password=unset
 spring.fc-datasource.url=unset
 spring.fc-datasource.username=unset
+spring.fc-datasource.minIdle=2
+spring.fc-datasource.maxPoolSize=10
+spring.fc-datasource.idleTimeout=60000
```
giangnguyen2, line 35: I kinda took the liberty of settimh minimum number of idle connection to 2, max number of connection to 10 and idle timeout to 1 minutes. If we think that we should set these values differently then I can change it

as-va (approved):

PR #300, opened 2022-02-17
API-12492 - Follow-on work to ensure that only Covid19 CMS detailed services allowed for V0

mmccollum-va:
Changes made to ensure that only Covid19 CMS detailed services are allowed for V0. This is follow-on to work currently completed in both API-12492 and API-12143.

### DIFF
--- a/facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/DeserializerUtil.java
+++ b/facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/DeserializerUtil.java
@@ -1,0 +1,444 @@
+package gov.va.api.lighthouse.facilities.api;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import lombok.NonNull;
+import lombok.experimental.UtilityClass;
+
+@UtilityClass
+public class DeserializerUtil {
+  /** Obtain active status node using snake_case or camelCase notation. */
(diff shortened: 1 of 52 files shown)


giangnguyen2:
Just curious. Wouldn't we be able to just filter out all services except for covid 19 based on service id when transform a `DatamartDetailedService` to V0 `DetailedService` in `DetailedServiceTransformerV0`

mmccollum-va:
Use of the various deserializers ensures the expected endpoint ingest behavior for V0/V1 CMS detailed services. For CMS detailed services uploaded into FAPI, the following update to `DetailedServiceTransformerV0` ensures that only Covid19 CMS detailed services will be allowed for V0:
```
  /** Transform a list of DatamartDetailedService to a list of version 0 DetailedService. */
  public static List<DetailedService> toDetailedServices(
      @Valid List<DatamartDetailedService> detailedServices) {
    return (detailedServices == null)
        ? null
        : !detailedServices.isEmpty()
            ? detailedServices.stream()
                .map(DetailedServiceTransformerV0::toDetailedService)
                .filter(
                    ds ->
                        ds.serviceId()
                            .equals(
                                uncapitalize(DatamartFacility.HealthService.Covid19Vaccine.name())))
                .collect(Collectors.toList())
            : new ArrayList<>();
  }
```

as-va (approved):

PR #303, opened 2022-02-24
API-13697 F: V1: Handle benefits services special case

mmccollum-va:
Datamart and versioned benefits services updated to properly handled creating `eBenefitsRegistrationAssistance` from `serviceId` or service name.


mmccollum-va:
- Use of deserializers found to be non-performant in this case
- Please refer to efforts found in [API-15279](https://github.com/department-of-veterans-affairs/lighthouse-facilities/pull/334) and [API-16147](https://github.com/department-of-veterans-affairs/lighthouse-facilities/pull/361)

PR #329, opened 2022-03-23
API-14307 F Improve CMS processing for V0 all endpoint

giangnguyen2:
Apply Covid19Vaccine service to facility health services list during CMS payload processing and facility daily reload


facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/CmsOverlayCollector.java
```diff
@@ -56,6 +71,16 @@ public static boolean containsCovidService(List<DatamartDetailedService> detaile
     return overlays;
   }
 
+  /** Return a map of facilities that have covid 19 vaccines. */
+  public HashMap<String, HealthService> loadCmsOverlayDetailedServices() {
```
as-va, line 75: I don't like this name for this function. The name implies that it is loading all overlay services, but it is actually only returning a small subset of services. It would be clearer if it were something more like `getCovid19VaccineServices`, `getV0DetailedServices`, `getServicesForFacilityStorage` or something more specific like that.

facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/CmsOverlayCollector.java
```diff
@@ -56,6 +71,16 @@ public static boolean containsCovidService(List<DatamartDetailedService> detaile
     return overlays;
   }
 
+  /** Return a map of facilities that have covid 19 vaccines. */
```
as-va, line 74: Can you add a small note to the comment along the lines of `This is a V0 utility function.`? There is only one set of collector files so this routine can't be isolated in one of the `*V0` or `v0/*` files; it has to be in the mainline. That will make it more difficult to find when we do the eventual retirement of V0 code. Having a note like that will make it easier to find with a `grep -i v0` in the future so we can identify this as code that can be removed.

facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/FacilitiesCollector.java
```diff
@@ -264,4 +269,28 @@ void updateOperatingStatusFromCmsOverlay(List<DatamartFacility> datamartFaciliti
...
+        facilityHealthServices.add(facilityCovid19Services.get(datamartFacility.id()));
+        if (datamartFacility.attributes().services().health() != null) {
+          facilityHealthServices.addAll(datamartFacility.attributes().services().health());
+        }
+        List<DatamartFacility.HealthService> facilityHealthServiceList =
+            new ArrayList<>(facilityHealthServices);
+        Collections.sort(facilityHealthServiceList);
+        datamartFacility.attributes().services().health(facilityHealthServiceList);
+      } else {
+        log.warn("No COVID-19 vaccines for facility: {}", datamartFacility.id());
```
as-va, line 292: Not sure this rises to the level of a warning. The _default_ case is no vaccine service present: there are ~300 facilities with this service out of ~2500 facilities in the API. This would create 2200 lines of warning each time it runs.

giangnguyen2: Agreed. This has been removed

facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/CmsOverlayCollector.java
giangnguyen2: Updated

as-va (approved):

PR #336, opened 2022-04-27
Removed pagination language from the /nearby endpoint documentation

ChrisBratti-VA:


mmccollum-va (dismissed):

as-va (changes requested):
This also needs to take out the Swagger Parameter annotations further down that add the `page` and `per_page` terms to the schema. Those are read by the Portal and put into the Curl Generator widget.

as-va (approved):

PR #339, opened 2022-05-04
Api 14991 consume health connect cms

ChrisBratti-VA:


facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/v0/CmsOverlay.java
```diff
@@ -22,4 +22,44 @@ public class CmsOverlay {
...
+  @Data
+  @Builder
+  @JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY)
+  @JsonInclude(value = Include.NON_NULL, content = Include.NON_NULL)
+  @Schema(description = "Information about the CMS system.", nullable = true)
+  public static final class SystemInformation {
+
+    @Schema(
+        description = "Health care system name",
+        example = "Example Health Care System Name",
```
giangnguyen2, line 39: Is this a valid example?

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
```diff
@@ -231,6 +231,13 @@ ResponseEntity<Void> deleteCmsOverlayById(
       }
       log.info("Deleting detailed_services node from overlay for id: {}", sanitize(id));
       overlayEntity.cmsServices(null);
+    } else if (thisNodeOnly.equalsIgnoreCase("system_information")) {
```
giangnguyen2, line 234: In the code block starting on line 216 when specific node is not specified, don't we need to delete `system information` as well?

ChrisBratti-VA: Good catch, yeah we should be deleting the system_information node in that block as well, making the update now

facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/v0/CmsOverlay.java
ChrisBratti-VA: The example should probably be updated, but I'm not 100% what to put as an example as I don't know what the real system names look like

facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/v0/CmsOverlay.java
```diff
@@ -22,4 +22,44 @@ public class CmsOverlay {
...
+  @Valid
+  @JsonProperty("system")
+  SystemInformation systemInformation;
+
+  @Data
+  @Builder
+  @JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY)
+  @JsonInclude(value = Include.NON_NULL, content = Include.NON_NULL)
+  @Schema(description = "Contains information pertaining to a CMS system.", nullable = true)
+  public static final class SystemInformation {
```
as-va, line 35: I'm thinking we should change this to `HealthCareSystem`. Just "SystemInformation" is a little too generic and could lead someone not already familiar with the design to assume this structure contains things like ram, cpu, etc.

facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/v0/CmsOverlay.java
```diff
@@ -22,4 +22,44 @@ public class CmsOverlay {
...
+
+  @Valid
+  @JsonProperty("system")
+  SystemInformation systemInformation;
+
+  @Data
+  @Builder
+  @JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY)
+  @JsonInclude(value = Include.NON_NULL, content = Include.NON_NULL)
+  @Schema(description = "Contains information pertaining to a CMS system.", nullable = true)
```
as-va, line 34: A Health Care System is a construct of the VHA, not just the CMS. Consumers would not care that CMS is our source for this information. Also, as with above comment would be better to spell out the full HCS. Something like `Describes the Health Care System for a facility.`

facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/v1/CmsOverlay.java
```diff
@@ -24,10 +25,47 @@ public class CmsOverlay implements CanBeEmpty {
...
     return (operatingStatus() == null || operatingStatus().isEmpty())
         && ObjectUtils.isEmpty(detailedServices());
   }
+
+  @Data
+  @Builder
+  @JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY)
+  @JsonInclude(value = Include.NON_NULL, content = Include.NON_NULL)
+  @Schema(description = "Contains information pertaining to a CMS system.", nullable = true)
+  public static final class SystemInformation {
```
as-va, line 43: See note on class name in V0 version of this file.

facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/v1/CmsOverlay.java
```diff
@@ -24,10 +25,47 @@ public class CmsOverlay implements CanBeEmpty {
...
   public boolean isEmpty() {
     return (operatingStatus() == null || operatingStatus().isEmpty())
         && ObjectUtils.isEmpty(detailedServices());
   }
+
+  @Data
+  @Builder
+  @JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY)
+  @JsonInclude(value = Include.NON_NULL, content = Include.NON_NULL)
+  @Schema(description = "Contains information pertaining to a CMS system.", nullable = true)
```
as-va, line 42: See note on description in V0 version of this file.

facilities/src/main/java/gov/va/api/lighthouse/facilities/CmsOverlayHelper.java
```diff
@@ -29,6 +29,14 @@ public static OperatingStatus getOperatingStatus(String operatingStatus) {
         : DATAMART_MAPPER.readValue(operatingStatus, OperatingStatus.class);
   }
 
+  /** Obtain DatamartFacility system informatoin from JSON string. */
```
as-va, line 32: _information_

facilities/src/main/java/gov/va/api/lighthouse/facilities/CmsOverlayHelper.java
```diff
@@ -42,4 +50,10 @@ public static String serializeDetailedServices(List<DatamartDetailedService> det
   public static String serializeOperatingStatus(OperatingStatus operatingStatus) {
     return (operatingStatus == null) ? null : DATAMART_MAPPER.writeValueAsString(operatingStatus);
   }
+
+  /** Obtain JSON string representation of DatamartFacility system informatoion. */
```
as-va, line 54: _information_

facilities/src/main/java/gov/va/api/lighthouse/facilities/CmsOverlayTransformerV1.java
```diff
@@ -23,6 +24,39 @@ public static DatamartCmsOverlay toVersionAgnostic(CmsOverlay overlay) {
...
                 overlay.detailedServices()))
+        .systemInformation(transformSystemInformation(overlay.systemInformation()))
         .build();
   }
+
+  /**
+   * Transform DatamartCmsOverlay.SystemInformation object to version 0 CmsOverlay.SystemInformation
+   * object
+   */
+  public static gov.va.api.lighthouse.facilities.api.v1.CmsOverlay.SystemInformation
```
as-va, line 35: Shouldn't need the fully qualified path here.

edit: or in the lines below.

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
```diff
@@ -231,6 +232,13 @@ ResponseEntity<Void> deleteCmsOverlayById(
...
```
as-va, line 235: This is an internal endpoint just for our use so we can have this work however we want - that said, I think this would be clearer to use the name of the node from the payload (i.e. `system`) rather than `system_information` here. In the case of the other nodes, the payload name and our internal name are the same. This is the first case where they differ, so we should define expectation here. My feeling is that deleting an overlay node should be reciprocal to creating the node, and `system` is the key that creates that portion of the overlay.

facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/v0/CmsOverlay.java
as-va: Make sure to also update other occurrences of the name in supporting classes such as Entities, Datamart, etc.

as-va (changes requested):

facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/v0/CmsOverlay.java
```diff
@@ -25,14 +25,14 @@ public class CmsOverlay {
 
   @Valid
   @JsonProperty("system")
-  SystemInformation systemInformation;
+  CmsOverlay.HealthCareSystem healthCareSystem;
```
as-va, line 28: This class definition is in this file; should not need the `CmsOverlay.` qualifier.

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalFacilitiesController.java
```diff
@@ -231,6 +232,13 @@ ResponseEntity<Void> deleteCmsOverlayById(
       }
       log.info("Deleting detailed_services node from overlay for id: {}", sanitize(id));
       overlayEntity.cmsServices(null);
+    } else if (thisNodeOnly.equalsIgnoreCase("system")) {
```
as-va, line 235: In the next `if` block below this section, you need to add a null check against this new field. This section goes through each major node type and sequentially checks if they need to be deleted, setting the fields to null if so. Afterwards, we update the record: if _all_ the nodes have been wiped out, we delete the entire overlay, otherwise we save the modified overlay to persist the fields that remain while blanking the null ones. Currently, since `healthCareSystem` is not checked in that calculation, if the other fields have been blanked but a system node remains, the overlay will be deleted and erroneously delete this node as well.

as-va (changes requested):

facilities-tests/src/test/java/gov/va/api/lighthouse/facilities/tests/CmsOverlayIT.java
```diff
@@ -426,6 +433,25 @@ void multiOverlayUpdate() {
     assertThat(facility.attributes().detailedServices())
         .usingRecursiveComparison()
         .isEqualTo(detailedServices());
+    ExpectedResponse.of(
+            requestSpecification()
+                .contentType("application/json")
+                .body(MAPPER.writeValueAsString(CmsOverlay.builder().healthCareSystem(sysInfo)))
```
giangnguyen2, line 439: Does this `CmsOverlay.builder().healthCareSystem(sysInfo)` work? I would think it needs `CmsOverlay.builder().healthCareSystem(sysInfo).build()`

ChrisBratti-VA: It actually does seem to work, I guess the mapper also supports the CmsBuilder object and serializes it to a String correctly. However I think it would be best practice to add the `.build()` on, making the update now

facilities/src/main/java/gov/va/api/lighthouse/facilities/CmsOverlayEntity.java
```diff
@@ -34,4 +40,18 @@ public class CmsOverlayEntity {
...
+  private String healthCareSystem;
+
+  @Default
+  @ElementCollection(targetClass = String.class)
+  @CollectionTable(
+      name = "cms_overlay_detailed_services",
+      schema = "app",
+      joinColumns = {@JoinColumn(name = "station_number"), @JoinColumn(name = "type")})
+  @Column(length = 48, name = "overlay_detailed_services")
+  private Set<String> overlayServices = new HashSet<>();
```
giangnguyen2, line 56: I think we should remove this attribute, This change is for API-14307 and it needs other changes in API-14307. It most likely will cause bad side effect during reload without other changes.

as-va: Agreed. I have not noticed any side effects in testing, but this does not need to be here and should stay in 14307 with the rest of this feature.

ChrisBratti-VA: I'm in agreeance as well, I didn't see any issues with removing it but wanted to get other opinions before removing it completely. Making that update now

as-va (approved):

PR #340, opened 2022-05-11
API-15796 - F: Add supplemental status support to operating status

mmccollum-va:
Addition of CMS supplemental status to operating status for a facility.


JamesDanielBell-VA (dismissed):

facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/v0/Facility.java
```diff
@@ -410,6 +410,31 @@ public static final class OperatingStatus {
...
+    @JsonProperty(required = true)
+    @Schema(description = "Unique id for supplemental status.", example = "COVID_HIGH")
+    String id;
+
+    @Valid
+    @NotNull
+    @JsonProperty(required = true)
+    @Schema(
+        description = "Descriptive label for supplemental status.",
+        example = "This is a high Covid level")
```
as-va, line 436: I just got an update from CMS with a sample of their current terminology. Could we change the examples for the sample status to these?
```
id: COVID_LOW
label: "COVID-19 health protection guidelines: Levels low"
```

as-va (approved):

PR #346, opened 2022-05-20
Api 14992 add health connect number to facility

ChrisBratti-VA:


facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/FacilitiesCollector.java
```diff
@@ -265,13 +265,15 @@ public void updateOperatingStatusFromCmsOverlay(List<DatamartFacility> datamartF
...
-                    : cmsOverlay.healthCareSystem().healthConnectPhone());
+        if (datamartFacility.attributes().phone() != null) {
+          datamartFacility
+              .attributes()
+              .phone()
+              .healthConnect(
+                  cmsOverlay.healthCareSystem() == null
+                      ? null
+                      : cmsOverlay.healthCareSystem().healthConnectPhone());
+        }
```
giangnguyen2, line 276: Even when datamartFacility.attributes().phone() == null meaning there is no `fax`, `main`, `pharmacy, afterHours`, `patientAdvocate`, `mentalHealth`, `enrollmentCoordinator` from VAST but there is a `healthCareSystem` from CMS, we should still create a phone attribute for the facility and set the `healthConnect` for it

ChrisBratti-VA: Good point, I'll update that conditional

giangnguyen2 (approved):

as-va (approved):

PR #348, opened 2022-05-26
Moved the processing of the operating status from the facilities requ…

ChrisBratti-VA:
…est to the reload process. Cleaned up existing tests


facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/FacilitiesCollector.java
```diff
@@ -286,6 +286,20 @@ public void updateOperatingStatusFromCmsOverlay(List<DatamartFacility> datamartF
       } else {
         log.warn("No cms overlay for facility: {}", datamartFacility.id());
       }
+      if (datamartFacility.attributes() != null) {
```
giangnguyen2, line 289: This check is not really necessary because the attributes is already marked as NoNull in `DatamartFacility `

ChrisBratti-VA: Removing that check ended up failing some unit tests like `FacilitiesCollectorTest.verifyResponse()`. I think you're right that in a real example the 'attributes()' would not be null, but I figured it was better to leave that check in then change the unit tests around this new functionality

giangnguyen2: If we need to keep it for test to pass then I'm okay with it

facilities-tests/src/test/java/gov/va/api/lighthouse/facilities/tests/CmsOverlayIT.java
```diff
@@ -534,6 +531,76 @@ void saveForUnknownFacility() {
...
+    assertThat(facility.attributes().operatingStatus().code())
+        .isEqualTo(OperatingStatusCode.CLOSED);
+    // Reload
+    ExpectedResponse.of(
+            requestSpecificationInternal()
+                .request(Method.GET, svcInternal.urlWithApiPath() + "internal/management/reload"))
+        .expect(200);
+    // After reload, ensure that the operating status was updated from CLOSED to NORMAL. Since we
+    // are testing with a
+    // vba facility, the activeStatus should always be null during the reload. In that case, the
```
JamesDanielBell-VA, line 590: If I understand right, if active_status is null then operating_status should be null after reload?
When I update active_status and operating_status to null in db and hit /reload looks like operating_status is being updated to "NORMAL"
Screen Shot 2022-06-09 at 5 46 53 PM

giangnguyen2 (approved):

facilities-tests/src/test/java/gov/va/api/lighthouse/facilities/tests/CmsOverlayIT.java
ChrisBratti-VA: My comment here was incorrect, it should be defaulting to NORMAL if the activeStatus is null, updated

giangnguyen2 (dismissed):

as-va (approved):

PR #360, opened 2022-06-30
Api 15738 process operating status during reload

ChrisBratti-VA:
Temporary solution to fix QA errors. IT tests will run a reload before running other tests to populate any null operating_statuses on facilities


as-va (approved):
Approving for a temporary patch.

PR #367, opened 2022-07-19
Fix to anticipate ATC name changes, this will allow for both services…

ChrisBratti-VA:
… names to be used


as-va:
This was an alternate attempt to maintain compatibility with both forms of ATC's data during their data switchover, but as their update is complete, this mitigation branch is no longer needed. These updates were captured in https://github.com/department-of-veterans-affairs/lighthouse-facilities/pull/366

PR #368, opened 2022-07-20
Api 17646 add snapshot of orthopedic service

ChrisBratti-VA:


facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/FacilitiesCollector.java
```diff
@@ -72,24 +74,24 @@ public FacilitiesCollector(
 
   /** Caregiver support facilities given a resource name. */
```
as-va, line 75: Comment needs updated since this function has been expanded.

ChrisBratti-VA: Good catch, just pushed update

facilities/src/main/resources/ortho_stations.txt
```diff
@@ -0,0 +1,221 @@
...
+691
+691A4
+691GB
+691GG
+691GL
+691GM
+693
+693B4
+695
+695GD
```
as-va, line 221: Somehow the last line of this file got cut off. The last line should be `756` for a total of 222 entries.

ChrisBratti-VA: Whoops I cut that off while testing and looks like I forgot to add it back in, updating now

as-va (approved):

PR #369, opened 2022-07-21
Api 17698 atc hotfix

ChrisBratti-VA:


facilities/src/main/java/gov/va/api/lighthouse/facilities/collector/HealthsCollector.java
```diff
@@ -170,7 +170,14 @@ private ListMultimap<String, AccessToCareEntry> loadAccessToCare() {
         log.warn("AccessToCare entry has null facility ID");
         continue;
       }
-      map.put(upperCase("vha_" + entry.facilityId(), Locale.US), entry);
+      if (entry.facilityId().equals("531")) {
+        if (entry.apptTypeName().equalsIgnoreCase("PRIMARY CARE")
+            && entry.sliceEndDate().equalsIgnoreCase("2022-07-18T00:00:00")) {
+          // Get it outta here
+        }
+      } else {
```
as-va, line 178: This won't work. This `else` is attached to the outermost `if`, so the inner `if` does nothing whether it is true or false while the outer `if` means that the map insert will only happen when the id != 531 and _no_ services for id 531 will get added.

as-va: You should either add a `continue` inside the inner `if` alongside the comment and get rid of this `else` block entirely, OR combine the two `if`s into one with all the conditions and-ed and leave the `else` as is.

ChrisBratti-VA: Ahhh good catch, I had broken it apart so I could catch a breakpoint there and forgot to put it back into one if statement, pushing out update now

as-va (approved):

PR #378, opened 2022-08-18
API-17481 F: Spike: Investigate how to handle service object data for V1

mmccollum-va:
- Decouple collection of ATC service data
   - Access-to-care decoupled from `FacilitiesCollector` - `HealthsCollector` - `HealthTransformer`
      - Individual collector  created for collecting ATC service data
   - Access-to-pwt has also been decoupled with an individual collector created for this data collection
- Sources of truth
   - CMS source of truth for V1 facility health services, even if service is present in ATC
      - For those facility health services present in ATC, but absent in CMS, ATC service is used
   - ATC source of truth for V0 facility health services, apart from Covid-19, which is sourced from CMS
- Source tracking changes have been incorporated to facilitate display of desired service name
- `last_updated` date field added to Datamart/V0/V1 detailed services to facilitate determination of most up-to-date CMS service name
- `changed` field removed from Datamart/V0/V1 detailed services
   - It was determined that this field was not in use by CMS and was removed on CMS side prior to completion of their push delivery system
- Changes made to ensure that when uploading overlay services, no buildup occurs within time span between reload occurrences of service objects containing previous service name value
- Changes made to ensure that for each service, latest service name update overall corresponding to that `serviceId` is utilized upon `reload`


mmccollum-va:
- [ ] Update merge target from `master` to `API-0000-v1-integration-phase-3` branch

JamesDanielBell-VA:
If I post payload with service name 'COVID-19' and hit `http://localhost:8085/v0/facilities/vha_561GF`, I can see the name change in the V0 response. I thought I recalled that V0 should be un-affected. Am I mistaken or is this the behavior we expect?
This is happening before and after `/reload`

image

JamesDanielBell-VA (dismissed):

PR #382, opened 2022-08-24
[API-18754] F: V1 Make service reporting configurable based on source

giangnguyen2:
Make service reporting configurable based on source:

- V0 considers all services with source = ATC, DST, BISL, internal and Covid19Vaccine which has source = CMS
- V1 considers whatever services with sources that are specified in application config file. 
- A single service should only show up once even if it comes from two sources.
- Now that all services regardless of their sources (ATC or CMS) are stored in the services attribute of the FacilityEntity, we only need to search for service in this attribute in JPA criteria. We don't need to look at both services attribute and overlayServices attribute in FacilityEntity.
- Now JPA ServicePredicate is built with a Set of String instead of a Set of ServiceType (change from Set<ServiceType> services to Set<String> services. This is to match the type of services in FacilityEntity. In addition, when a consumer sends a request with a service id, we need to generate a Set of service string with all appropriate permutation of sources depending on the version.


facilities/src/main/java/gov/va/api/lighthouse/facilities/FacilityRepository.java
```diff
@@ -48,49 +42,15 @@
...
-          services.stream()
-              .map(
-                  svc ->
-                      criteriaBuilder.isMember(
-                          capitalize(svc.serviceId()), root.get("overlayServices")))
-              .toArray(Predicate[]::new);
-      Predicate anyOverlayService = criteriaBuilder.or(overlayServicePredicates);
-
-      return criteriaBuilder.or(anyFacilityService, anyOverlayService);
+      return criteriaBuilder.or(servicePredicates);
```
giangnguyen2, line 53: A note here: Now that all services regardless of their sources (ATC or CMS) are stored in the services attribute of the FacilityEntity, we only need to search for service in this attribute in JPA criteria.

PR #389, opened 2022-09-10
API-19608 - F: V0: CMS overlay upload of custom services

mmccollum-va:
CMS utilizes `service_api_id`, rather than `service_id` or `serviceId`, as unique service identifier. JsonAliases corrected/updated for Datamart and versioned DetailedServices to enable correlation of `service_id`, `serviceId`, and `service_api_id` JSON fields to respective `serviceId` class attribute.


as-va (approved):

as-va (approved):
These changes add over 12 minutes to any build that executes the CmsOverlayIT tests due to the reloads introduced there. Will approve this in order to get the code fix deployed but we should revisit these tests for refactoring in the near future.

PR #397, opened 2022-10-03
Api 20221 search by address error response

ChrisBratti-VA:
Removed `/nearby` search by address functionality, now returning a deliberate 400 error.

Added logic in the `WebExceptionHandler` to catch and strip the search by address parameters from the error message produced by incorrect parameter combinations to prevent users from thinking its a valid query, ex -

`Parameter conditions "lat, lng" OR  "street_address, city, state, zip" not met for actual request parameters: foo={bar}` - Old message

`Parameter conditions "lat, lng" not met for actual request parameters: foo={bar}` - New message

Updated the portal documentation to reflect this functionality has been deprecated


ChrisBratti-VA:
New PR has been opened for this effort ->
https://github.com/department-of-veterans-affairs/lighthouse-facilities/pull/438#issue-1674989988

PR #399, opened 2022-10-11
API-20452 - F: Spike: Investigate improvements of api level processing for the /nearby endpoint

mmccollum-va:
- Parallel Stream Optimization - use `findAny()`, rather than `findFirst()`, when using parallel search based on `serviceId`. `serviceId` guaranteed to be unique across all services. `findFirst` causes parallel to continue search to the left to ensure match is first after match is found.
- Updates made to prevent `BufferUnderflowException` when decoding path from compressed band data


bschofield-va (changes requested):
Preventing merge into master

facilities-api/src/test/java/gov/va/api/lighthouse/facilities/api/pssg/PathEncoderTest.java
```diff
@@ -62,60 +63,62 @@ void roundTrip() {
...
-    assertThat(coords[0]).isCloseTo(40.444, offset(0.01));
-    assertThat(coords[1]).isCloseTo(-10.111, offset(0.01));
-    iter.next();
-    assertThat(iter.currentSegment(coords)).isEqualTo(PathIterator.SEG_LINETO);
-    assertThat(coords[0]).isCloseTo(20.222, offset(0.01));
-    assertThat(coords[1]).isCloseTo(-10.111, offset(0.01));
-    iter.next();
-    assertThat(iter.currentSegment(coords)).isEqualTo(PathIterator.SEG_CLOSE);
+    Optional<Path2D> path = PathEncoder.create().decodeFromBase64(path64);
+    if (path.isPresent()) {
```
bschofield-va, line 67: From a test perspective... aren't we in control of whether we expect the `path` to be present or not? I'm not sure I understand why this test may not perform any assertions at all.

facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/pssg/PathEncoder.java
```diff
@@ -89,35 +90,38 @@ private Path2D decode(byte[] compressPath) {
...
-      for (int coordNumber = 0; coordNumber < numberOfCoordinates; coordNumber++) {
-        int scaledLong = buffer.getInt();
-        int scaledLat = buffer.getInt();
-        double longitude = unscale(scaledLong);
-        double latitude = unscale(scaledLat);
-        if (coordNumber == 0) {
-          path.moveTo(longitude, latitude);
-        } else {
-          path.lineTo(longitude, latitude);
+    if (buffer.hasRemaining()) {
```
bschofield-va, line 93: I get that is method is handling for the buffer underflow condition that can arise from invalid path data. _How does that path data become corrupted to begin with?_ I can certainly be wrong, but this feels like this is compensating for a symptom of a problem rather than fixing the root problem itself.

facilities-api/src/main/java/gov/va/api/lighthouse/facilities/api/v0/Facility.java
```diff
@@ -80,7 +80,7 @@ public static Optional<BenefitsService> fromServiceId(String serviceId) {
       return Arrays.stream(values())
           .parallel()
           .filter(bs -> bs.serviceId().equals(serviceId))
-          .findFirst();
+          .findAny();
```
bschofield-va, line 83: out of curiosity, does `findAny` give us better performance than `findFirst`?

mmccollum-va: In theory, yes. Intent is to quickly find services based on their unique identifier. Both `findAny` and `findFirst` can achieve this using parallel streams, however, `findFirst` must take additional steps to ensure match is first in collection. `findAny` used here to avoid the extra steps.

facilities-api/src/test/java/gov/va/api/lighthouse/facilities/api/pssg/PathEncoderTest.java
mmccollum-va: This change was made to account for DTBs with no geometries

mmccollum-va:
- GitHub [PR](https://github.com/department-of-veterans-affairs/lighthouse-facilities/pull/399) has been open for **`2.5` months**
    - It has been recommended to close this [PR](https://github.com/department-of-veterans-affairs/lighthouse-facilities/pull/399) and re-open it if it becomes useful and/or needed in the future

PR #437, opened 2023-04-14
Api 25577 cms data model updates

ChrisBratti-VA:


facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalCmsOverlayTransformerController.java
```diff
@@ -0,0 +1,54 @@
...
+    ArrayList<String> updatedIds = new ArrayList<>();
+    entities.forEach(
+        cmsOverlayEntity -> {
+          try {
+            String newCmsServices =
+                MAPPER.writeValueAsString(
+                    CmsOverlayHelper.getDetailedServices(cmsOverlayEntity.cmsServices()));
+            if (!cmsOverlayEntity.cmsServices().equalsIgnoreCase(newCmsServices)) {
+              cmsOverlayEntity.cmsServices(newCmsServices);
+              updatedIds.add(cmsOverlayEntity.id().toIdString());
```
as-va, line 44: I'm not a huge fan of updating collections outside of a stream from within a stream like this as it can lead to difficult to track down bugs when the stream is parallelized. `ArrayList` is not thread-safe so multiple threads trying to add to it can cause a bad result. Given that this code is serial at the moment _and_ this function will be short lived and removed soon after deploying I'm ok leaving this as is. But keep in mind for future coding that if this were to be a long-lived function, even though this would work now, someone could refactor this to a parallelstream in the future to try to improve performance and suddenly this code would be problematic. Writing streams to be thread-safe even when only intended to be run in serial (when possible) is good practice to protect against this type of silent bug introduction.

facilities/src/test/java/gov/va/api/lighthouse/facilities/CmsOverlayTransformerV0Test.java
```diff
@@ -198,16 +199,17 @@ private DetailedService getDetailedService(
...
         .appointmentLeadIn(
             "Your VA health care team will contact you if you???re eligible to get a vaccine "
                 + "during this time. As the supply of vaccine increases, we'll work with our care "
                 + "teams to let Veterans know their options.")
-        .onlineSchedulingAvailable("true")
+        .onlineSchedulingAvailable(true)
         .serviceLocations(
             List.of(
                 DetailedService.DetailedServiceLocation.builder()
+                    .officeName("ENT Clinic")
```
as-va, line 212: Is there also a test point where `officeName` is not provided (`null`)?

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalCmsOverlayTransformerController.java
as-va, line 1: Can you add a comment in here somewhere visible that this endpoint is only intended to be used once and that this entire file can be removed after the deployment containing 25577? That will remove ambiguity in the future to anyone who sees this file that it does not need to be maintained going forward after we verify that the database is properly updated to the new data format.

facilities-tests/src/test/java/gov/va/api/lighthouse/facilities/tests/v0/CmsOverlayIT.java
```diff
@@ -1092,10 +1105,90 @@ void updateFacilityOperatingStatus() {
...
+        ExpectedResponse.of(
+                requestSpecification()
+                    .request(
+                        Method.GET,
+                        svc.urlWithApiPath() + "v0/facilities/" + facilityID + "/cms-overlay"))
+            .expect(200)
+            .expectValid(gov.va.api.lighthouse.facilities.api.v0.CmsOverlayResponse.class);
+    assertThat(
+            cmsOverlay.overlay().detailedServices().get(0).serviceLocations().get(0).officeName())
+        .isEqualTo("ENT Clinic");
```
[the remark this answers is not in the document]
as-va: This actually doesn't need to be in the ITs, it's probably better served as a unit test data point but I didn't see a similar test done there either.

as-va (changes requested):

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalCmsOverlayTransformerController.java
ChrisBratti-VA: What's funny is I _did_ updat this before opening the PR in [the last commit](https://github.com/department-of-veterans-affairs/lighthouse-facilities/pull/437/commits/e5bfa69102db5d51ddd14f9857eb7b78c2afa79e) but for some reason don't see it reflected here.

ChrisBratti-VA: That's a good point. What would be the best alternative in your eyes in a scenario like this? A different data structure like a synchronized list or a different approach to how we would load the list?

facilities-tests/src/test/java/gov/va/api/lighthouse/facilities/tests/v0/CmsOverlayIT.java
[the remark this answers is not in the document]
ChrisBratti-VA: Are you saying that this existing test as well as the new test should both be unit tests instead of ITs, or just the new test?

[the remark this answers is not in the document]
ChrisBratti-VA: Ah okay that makes more sense. And I agree, I wasn't 100% sure on this IT test since, as you said, it is covered by the other IT tests implicitly, but I didn't want it to seem like I wasn't testing the `office_name` field. I think the unit tests are also set up to test these fields so it should be caught by those as well, but I can add one or two in there any way if that would be best practice

facilities/src/main/java/gov/va/api/lighthouse/facilities/InternalCmsOverlayTransformerController.java
as-va: Generally speaking I'd try to structure the logic to avoid side effects on objects outside of the stream itself: doing things like `.map()`-ing to pull out the field of interest, `.filter()` to get only the relevant ones and `.collect()`ing to a list as the terminator of the stream. If that's not possible (or inconvenient), it'd be a judgement call on using well defined side effects, changing types to synchronized variants, etc based on what makes the code easiest to read and maintain (while also considering performance concerns when applied in a critical processing path).

as-va:
This PR was superseded by #443.

